PR #1, opened 2023-03-14
implement 1bpp fonts in console.c, shrink default_font.bin to match

asiekierka:
This PR implements the following:

* Loading 1BPP fonts in console.c. The code clearly implies this was intended functionality, but for some reason it hasn't been implemented. This probably adds some space, but it seems to be negligible (<1KB for sure) on a brief glance.
* Replacing default_font.bin:
    * The font is now derived from a copy of Viznut's public-domain [Unscii](http://viznut.fi/unscii/) font, a personal go-to of mine. (We can pick a different 8x8 font, though - the sky's the limit and you may have better taste.)
    * The font is now 1BPP instead of 4BPP-with-only-colors-0-and-1-used, saving 6KB of space.
    * The font now removes characters 128-255, which were never loaded anyway by the code (see console.c definition), saving an additional 1KB of space.
    * The font now only provides characters 32-126 (ASCII), excluding the drawing characters 0-31 and 127; this saves an additional 264 bytes of space.

Of course, I'm willing to negotiate.


source/arm9/console.c
```diff
@@ -414,23 +411,37 @@ void consoleLoadFont(PrintConsole* console) {
...
-					temp |= 0xF00;
-				if(console->font.gfx[i] & 0xF000)
-					temp |= 0xF000;
-
-				console->fontBgGfx[i] = temp;
+			if (console->font.bpp == 1) {
+				for (i = 0; i < console->font.numChars * 8; i++) {
+					u8 row = ((const u8*) console->font.gfx)[i];
+					u32 temp = 0;
+					if(row & 0x80) temp |= 0xF;
```
asiekierka, line 418: The order may appear reversed here. This is to match the typical bit order of PC/DOS character sets, as well as the 1bpp tile editing tools I've encountered such as YY-CHR.

PR #2, opened 2023-03-15
keyboard improvements

asiekierka:
* Use 16-bit instead of 32-bit keymap indices.
  * The only used indices, in practice, are a small set of values < 0 (DVK_* defines) and values 0-255 (the range of a `char`). Given that the DS's bottom display doesn't even have 65536 pixels total, 16-bit should be more than enough.
  * This saves 640 bytes of ROM space for the default keyboard.
* Remove unused key buffer if NO_DEVOPTAB is defined.
  * The key buffer is only used if `keyboardRead` is present, which is not used alongside libsysnds (which prefers `keyboardUpdate`). As such, it is removed, removing backspace functionality in edge cases but saving ~260 bytes of RAM.
* Resolve bounds check TODO, and fix a bug in handling non-32-wide keymaps while at it.

### DIFF
--- a/include/nds/arm9/keyboard.h
+++ b/include/nds/arm9/keyboard.h
@@ -64,7 +64,7 @@ typedef enum
 typedef struct KeyMap {
 	const u16* mapDataPressed;	/*!< the map for keys pressed*/
 	const u16* mapDataReleased;	/*!< the map for keys released */
-	const int* keymap; 			/*!< the lookup table for x y grid location to corresponding key */
+	const s16* keymap; 			/*!< the lookup table for x y grid location to corresponding key */
 	int width; 					/*!< width of the keyboard in grid spaces */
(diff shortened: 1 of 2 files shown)


include/nds/arm9/keyboard.h
```diff
@@ -162,7 +162,7 @@ void keyboardHide(void);
 	\param y the pixel y location
 	\return the key pressed or NOKEY if user pressed outside the keypad
 */
-int keyboardGetKey(int x, int y);
+s16 keyboardGetKey(int x, int y);
```
asiekierka, line 165: Changing function definitions is probably unnecessary, but it is more consistent. What do you think?

AntonioND: I think changing them is more honest, so that people don't use 4 bytes when they can get away with using 2.

PR #4, opened 2023-03-19
Remove frame-delay

NightScript370:
Comes with the benefit of not handling an un-needed variable! 🎉
Backport of https://github.com/DS-Homebrew/TWiLightMenu/commit/98982d74826b490353e34ba3b683ca915bc9c149

### DIFF
--- a/source/arm7/input.c
+++ b/source/arm7/input.c
@@ -14,7 +14,6 @@ enum{
 
 void inputGetAndSend(void){
 
-	static bool penDown = false;
 	static int sleepCounter = 0;
 
 	touchPosition tempPos = {0};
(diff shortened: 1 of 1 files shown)


AntonioND:
I'm a bit worried about this patch because of one of the comments in GBATEK:

> The TSC values may be garbage if the screen becomes newly pressed or released, to avoid invalid inputs: read TSC values at least two times, and ignore BOTH positions if ONE position was invalid.

This frame delay seems to have been added for this. Are you sure that it isn't a problem?

asiekierka:
Alternatively, though, this could be related to the multiple-read code found [here](https://github.com/blocksds/libnds/blob/master/source/arm7/touch.c#L149-L213).

AntonioND:
Ah, good catch, the documentation mentions "read twice", not during two frames. That looks more like the code meant to deal with the issue.

PR #9, opened 2023-03-19
Fix several EEPROM bugs

Epicpkmn11:
I see you already merged 9cb70adbe72222db8afc25bfc2ac5e24c2f4c8c0, but not this yet so figured I'd open a PR so I could add some notes

Original GodMode9i PRs:
- Line 82: DS-Homebrew/GodMode9i#88 -- fixes dumping 128 KiB EEPROMs (specifically Pokémon Mystery Dungeon: Explorers of Sky)
- Lines 100-181: DS-Homebrew/GodMode9i#79 -- fixes incorrect size detection when the first word was `00000000` because it inverted the number and treated `FFFFFFFF` as a stop condition iirc
   - Improved in: DS-Homebrew/GodMode9i#94 -- improves on the previous fix by checking two separate values, to remove the chance of a save ever coincidentally perfectly matching our check value

### DIFF
--- a/source/common/cardEeprom.c
+++ b/source/common/cardEeprom.c
@@ -79,7 +79,7 @@ int cardEepromGetType(void) {
 	if (( sr == 0xff && id == 0xffffff) || ( sr == 0 && id == 0 )) return -1;
 	if ( sr == 0xf0 && id == 0xffffff ) return 1;
 	if ( sr == 0x00 && id == 0xffffff ) return 2;
-	if ( id != 0xffffff) return 3;
+	if ( id != 0xffffff || ( sr == 0x02 && id == 0xffffff )) return 3;
 	
 	return 0;
 }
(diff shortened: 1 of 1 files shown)


source/common/cardEeprom.c
```diff
@@ -150,8 +163,22 @@ uint32 cardEepromGetSize() {
...
+				break;
+			case 0x2017:
+				return 8*1024*1024;		//	64Mbit(8 meg)
+				break;
+			}
+		}
+
+		if (id == 0xffffff) {
+			int sr = cardEepromCommand(SPI_EEPROM_RDSR);
+			if (sr == 2) { // Pokémon Mystery Dungeon - Explorers of Sky
```
asiekierka, line 179: Personal pet peeve, but naming games in the official repository feels wrong.

AntonioND, line 179: I think we can remove this comment about Pokemon Mystery Dungeon.

AntonioND: Yes, I think the comment should be removed.

Epicpkmn11: Removed

source/common/cardEeprom.c
```diff
@@ -97,22 +97,35 @@ uint32 cardEepromGetSize() {
...
+				cardWriteEeprom(0,(u8*)&buf4,4,type);
+
+				// Check if it still matches
+				cardReadEeprom(size,(u8*)&buf2,4,type);
+				if (buf2 == buf4) break;
+
+				// False match, write "TEST" back and keep going
+				cardWriteEeprom(0,(u8*)&buf3,4,type);
+			}
 			size += 8192;
```
asiekierka, line 124: Since we're here - are there non-power-of-two EEPROMs? Couldn't we hypothetically replace this with a slightly faster "size <<= 1"?

source/common/cardEeprom.c
```diff
@@ -97,22 +97,35 @@ uint32 cardEepromGetSize() {
...
-		} else {
-			buf3 = buf1;
-		}
+
+		// Write "TEST" to it
+		cardWriteEeprom(0,(u8*)&buf3,4,type);
+
+		// Loop until the EEPROM mirrors and the first word shows up again
 		int size = 8192;
-		while (1) {
```
asiekierka, line 109: Might want to put a size cap here, or inside the loop. Don't know what value would be sensible, but in theory one could make a hostile card implementation that causes an infinite loop.

asiekierka:
Too late; if you'd like to provide another PR, that's cool with me, but giving answers is fine too - I might look into it later.

source/common/cardEeprom.c
AntonioND: Ah, I didn't see your comment before merging. Yes, it would be nice to have a cap.

PR #10, opened 2023-03-19
More EEPROM improvements

Epicpkmn11:
Per @asiekierka's comments after #9 was merged, I've added a size limit of 8 MB (the largest DS EEPROM save, Art Academy) and changed it to only check powers of two. Per [melonDS's save DB](https://github.com/melonDS-emu/melonDS/blob/master/src/ROMList.h) non-power of 2 EEPROMs do not exist.

I've tested a handful of games using GM9i and it appears to work as before.


asiekierka:
LGTM

PR #15, opened 2023-03-25
fatfs: Implement re-entrancy protection.

asiekierka:
This will avoid filesystem corruption if two cothreads try to write to the filesystem at the same time.

### DIFF
--- a/include/nds/cothread.h
+++ b/include/nds/cothread.h
@@ -100,6 +100,13 @@ void cothread_yield_irq_aux(uint32_t flags);
 // Returns ID of the thread that is running currently.
 cothread_t cothread_get_current(void);
 
+// Initialize a mutex.
+static inline bool comutex_init(comutex_t *mutex)
+{
+    *mutex = 0;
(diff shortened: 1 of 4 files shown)


AntonioND:
I think it's actually worse than this. If I use DLDI to read, and I try to read from the DSi SD, it's possible that the threads will read the results in the wrong order because there is no ordering enforced for the FIFO messages. We should have just one lock for everything for now.

asiekierka:
Well, then I can `#define FF_FS_LOCK 1` for now; this will only allow one open file to exist at the same time.

AntonioND:
Ok, set it to 1 for now, we can make it a bit more fine-grained later.

asiekierka:
Except that would break programs which open more than one file in the same thread, and I wouldn't be surprised if such programs exist.

From internal discussions, it seems the best solution to FIFO ordering issues, at least for now, is to write a per-channel lock for thread-yielding FIFO set->get call pairs.

PR #16, opened 2023-03-25
fifo: Use locks to protect FIFO communications

AntonioND:
After introducing cothreads, whenever we yield from the ARM9 while waiting for the ARM7 to reply, it is possible that two threads get their responses in the wrong order.

Thread 1 sends message 1, it yields while waiting for the response, thread 2 sends message 2, and response 1 is ready, so it reads it without yielding, and eventually thread 1 sees response 2.

This PR introduces fifoWaitValueAsynch32() so that not all FIFO communications need to be asynchronous, only the communications that explicitly use this function.

It also switches the storage and sound FIFO subsystems to use locks and the new asynchronous wait function.

### DIFF
--- a/include/nds/fifocommon.h
+++ b/include/nds/fifocommon.h
@@ -335,20 +335,21 @@ int fifoGetDatamsg(int channel, int buffersize, u8 * destbuffer);
 	\param channel the channel to check.
 
 */
-//---------------------------------------------------------------------------------
-static inline void fifoWaitValue32(int channel) {
-//---------------------------------------------------------------------------------
 
-	while(!fifoCheckValue32(channel)) {
-#ifdef ARM9
-		cothread_yield_irq(IRQ_FIFO_NOT_EMPTY);
-#else
-		swiIntrWait(1,IRQ_FIFO_NOT_EMPTY);
-#endif
-	}
-
+static inline void fifoWaitValue32(int channel)
+{
+	while (!fifoCheckValue32(channel))
+		swiIntrWait(1, IRQ_FIFO_NOT_EMPTY);
 }
 
+static inline void fifoWaitValueAsync32(int channel)
+{
+	while (!fifoCheckValue32(channel))
+		cothread_yield_irq(IRQ_FIFO_NOT_EMPTY);
+}
+
+void fifoMutexAcquire(int channel);
+void fifoMutexRelease(int channel);
 
 #ifdef __cplusplus
 };
--- a/source/arm9/sound.c
+++ b/source/arm9/sound.c
@@ -47,12 +47,17 @@ int soundPlayPSG(DutyCycle cycle, u16 freq, u8 volume, u8 pan){
 	msg.SoundPsg.volume = volume;
 	msg.SoundPsg.pan = pan;
 
+	fifoMutexAcquire(FIFO_SOUND);
+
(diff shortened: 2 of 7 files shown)


source/arm9/sound.c
```diff
@@ -47,12 +50,17 @@ int soundPlayPSG(DutyCycle cycle, u16 freq, u8 volume, u8 pan){
...
 	msg.SoundPsg.pan = pan;
 
+	comutex_acquire(&sound_fifo_mutex);
+
 	fifoSendDatamsg(FIFO_SOUND, sizeof(msg), (u8*)&msg);
+	fifoWaitValueAsync32(FIFO_SOUND);
+	int result = fifoGetValue32(FIFO_SOUND);
 
-	while(!fifoCheckValue32(FIFO_SOUND));
+	comutex_acquire(&sound_fifo_mutex);
```
asiekierka, line 59: Shouldn't this be `comutex_release`?

source/arm9/sound.c
```diff
@@ -122,6 +138,8 @@ MicCallback micCallback = 0;
 void micBufferHandler(int bytes, void* user_data){
 	FifoMessage msg;
 
+	// TODO: Do we need a lock here?
```
asiekierka, line 141: This is inside an IRQ! Don't use yielding here.

include/nds/fifocommon.h
```diff
@@ -335,20 +335,18 @@ int fifoGetDatamsg(int channel, int buffersize, u8 * destbuffer);
...
-#endif
-	}
 
+static inline void fifoWaitValue32(int channel)
+{
+	while (!fifoCheckValue32(channel))
+		swiIntrWait(1, IRQ_FIFO_NOT_EMPTY);
 }
 
+static inline void fifoWaitValueAsync32(int channel)
```
asiekierka, line 345: It might be good to add equivalent Wait/WaitAsync functions for Address and Datamsg too.

source/arm9/sound.c
AntonioND: Ok, I'll change the note.

source/arm9/sound.c
AntonioND: Yeah, I've made the same mistake all over the file...

include/nds/fifocommon.h
AntonioND: That's fair, but I think that can be added outside of this PR, it's a bit complicated as it is now.

source/arm9/sound.c
```diff
@@ -122,6 +135,8 @@ MicCallback micCallback = 0;
 void micBufferHandler(int bytes, void* user_data){
 	FifoMessage msg;
 
+	// This is inside an IRQ, no locks required.
```
asiekierka, line 138: That's not what I meant! fifoWaitValueAsync32 will call cothread_yield, and I'm not sure if our yielding routine handles cothread_yield being called from inside an IRQ. My proposal was to keep this one blocking for the time being.

include/nds/fifocommon.h
asiekierka: OK

AntonioND:
Oh, I've changed the message to "fifo" instead of "storage" so GitHub hasn't detected the merge. Anyway, this is the commit https://github.com/blocksds/libnds/commit/942bc31eb480c0aba7c8f92d6a4391b02a649334

PR #17, opened 2023-03-26
peripheral: minor fixes

asiekierka:
* Fix a bug in which the Guitar Grip detection code would unexpectedly move Slot-1 control and/or main memory priority to ARM7.
* Improve rumble detection code. People from other communities have reported that the code for detecting a DS Rumble Pak isn't bulletproof, so it now matches GBATEK's recommended approach.
* Make rumble detection code "lazy" (as the for loop takes a few milliseconds).
* Add rumbleGetType/rumbleSetType - our code is known to not cover all auto-detection cases, and I don't have a comprehensive list right now, if ever.
* Add untested support for the DS rumble cartridge bundled with "Magukiddo". No detection for the time being, as I don't have a cart to test with.
* Add EXMEMCNT defines, remove private EXMEMCNT defines from `paddle.c`


include/nds/arm9/rumble.h
```diff
@@ -32,14 +32,21 @@
...
 typedef enum {
-   RUMBLE,
-   WARIOWARE
-}RUMBLE_TYPE;
+   RUMBLE_TYPE_UNKNOWN,
+   RUMBLE_TYPE_NONE,
+   RUMBLE_TYPE_PAK, /* DS Rumble Pak */
+   RUMBLE_TYPE_GBA, /* rumble included as part of GBA game cartridges */
+   RUMBLE_TYPE_MAGUKIDDO /* rumble/sensor cartridge bundled with Magukiddo */
+} RUMBLE_TYPE;
```
asiekierka, line 45: Breaking these definitions is probably fine, as they were never exposed to user code before anyway.

PR #26, opened 2023-04-08
sleep: Add ability for ARM9 to disable sleep

Lorenzooone:
Quick commit to make it easy to disable sleep from ARM9.


AntonioND:
Huh, it's weird this hadn't been implemented before.

PR #32, opened 2023-04-21
fatfs: Set the CWD to the path of the NDS file

AntonioND:
Previously, when initializing FAT, the default path was set to the root of the drive that contains the NDS file.

With this patch, fatInit() will also set the current working directory to the folder that contains the NDS file. This way it behaves more like it would in a real OS


AntonioND:
I've modified this code based on our chat (I've added a function to get the default CWD, and I use it during the initialization of FAT).

https://github.com/blocksds/libnds/commit/c2ac2e1d9ac4ebe1af517e68ba62d24670800732

PR #33, opened 2023-04-21
image, pcx, fatfs: minor fixes

asiekierka:
* image, pcx: fix missing checks, use bools, unify error checking approach - I spotted this one by looking at the commit history; just a minor improvement.
* fatfs: inline ff_memalloc/ff_memfree - they're just wrappers to malloc/free, so why not?
* fatfs: fix memory leak in fatInit() - these two issues have been spotted by using GCC 12's `-fanalyzer` static analyzer.

### DIFF
--- a/source/arm9/image.c
+++ b/source/arm9/image.c
@@ -12,11 +12,11 @@
 #include <nds/dma.h>
 #include <nds/ndstypes.h>
 
-void image24to16(sImage *img)
+bool image24to16(sImage *img)
 {
     u16 *temp = malloc(img->height * img->width * 2);
     if (temp == NULL)
-        return;
+        return false;
 
     for (int y = 0; y < img->height; y++)
     {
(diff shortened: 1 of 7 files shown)


source/arm9/image.c
```diff
@@ -83,21 +87,23 @@ void image8to16trans(sImage *img, u8 transparentColor)
...
     img->image.data16 = temp;
+
+    return true;
 }
 
-void imageTileData(sImage *img)
+bool imageTileData(sImage *img)
 {
     // Can only tile 8 bit data that is a multiple of 8 in dimention
-    if (img->bpp != 8 || (img->height & 3) != 0 || (img->width & 3) != 0)
```
asiekierka, line 91: Also, this check seems to have been entirely wrong?

AntonioND: It looks like it.

PR #37, opened 2023-09-14
Make it easier to fully map DS accessory memory in DSi mode

Lorenzooone:
This PR makes it so it's easier to map the 0x8000000 range of memory in DSi mode from regular code.
Before, it would end up in a different region depending on the platform (unless you un-mapped the switchable IWRAM and then re-mapped it. But that would mean un-mapping and re-mapping the DSi region at 0xC000000, so it would be an annoying process all around).

While it does impact the logical order, it makes it so the memory regions are more compact based on their address.

### DIFF
--- a/source/arm9/system/mpu_setup.s
+++ b/source/arm9/system/mpu_setup.s
@@ -20,14 +20,14 @@
 // ----+------------+--------+-----------+--------+-------+----+-----------------------
 //   2 | 0x00000000 |   4 KB | All       |  R/W   |   N   | N  | Alternate vector base
 // ----+------------+--------+-----------+--------+-------+----+-----------------------
-//   3 | 0x08000000 | 128 MB | DS, DSd   |  R/W   |   N   | N  | DS Accessory (GBA Cart)
+//   3 | 0x02000000 |  16 MB | DS    [1] |  R/W   |   N   | N  | DS non-cacheable main RAM
+//     | 0x02800000 |   8 MB | DSd       |        |       |    |
 //     | 0x03000000 |   8 MB | DSI, DSId |        |       |    | DSi switchable IWRAM
 // ----+------------+--------+-----------+--------+-------+----+-----------------------
 //   4 | 0x01000000 |  32 KB | All       |  R/W   |   N   | N  | ITCM
 // ----+------------+--------+-----------+--------+-------+----+-----------------------
-//   5 | 0x02000000 |  16 MB | DS    [1] |  R/W   |   N   | N  | Non-cacheable main RAM
-//     | 0x02800000 |   8 MB | DSd       |        |       |    |
-//     | 0x0C000000 |  16 MB | DSI       |        |       |    |
+//   5 | 0x08000000 | 128 MB | DS, DSd   |  R/W   |   N   | N  | DS Accessory (GBA Cart)
+//     | 0x0C000000 |  16 MB | DSI       |        |       |    | DSi non-cacheable main RAM
 //     | 0x0C000000 |  32 MB | DSId      |        |       |    | DSi debugger extended IWRAM
 // ----+------------+--------+-----------+--------+-------+----+-----------------------
 //   6 | 0x02000000 |   4 MB | DS        |  R/W   |   Y   | Y  | Cacheable main RAM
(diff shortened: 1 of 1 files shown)


AntonioND:
I disagree with this change. It makes the code more confusing and less maintainable, and the only benefit is that you can do something that isn't a good idea. Yesterday, switching memory regions was trivial because I only had to swap around a few numbers. After this commit, I would have to move more things to do something similar, and that's just asking for trouble.

The whole point of the memory regions in the MPU is to crash if the CPU accesses memory addresses that don't make sense, because most likely it's a programming error. Mapping 0x8000000 in DSi mode doesn't make sense, there is nothing there: https://problemkaputt.de/gbatek.htm#dsiiomap so it's a good thing that accessing that memory range causes exceptions.

If your application accesses the GBA slot, you can just use ``isDSiMode()`` to detect if you're running in DSi mode, and avoid reading from the slot: https://github.com/blocksds/libnds/blob/ea6f7c9a106d907acbd7c0f32a63bdaf8be388d7/include/nds/system.h#L92

And the final thing, if you choose to ignore the ``isDSiMode()`` function. Don't rely on the order of memory regions specified in that file. It's code that is internal to libnds, and I won't make any guarantee about preserving the current behaviour. It may change if there is a good reason for it (as it changed yesterday with commits https://github.com/blocksds/libnds/commit/25b1bb810afb86e595f7210493c459f4d2fb0f3d and https://github.com/blocksds/libnds/commit/ea6f7c9a106d907acbd7c0f32a63bdaf8be388d7). Any code that expects region X to refer to Y type of RAM may eventually break. If you really want to map memory regions in a reliable way, just copy the whole function to your application and call it from there at the beginning of ``main()``.

asiekierka:
Why would one map 0x8000000 in DSi mode, anyway? What's the developer story here?

Lorenzooone:
> Why would one map 0x8000000 in DSi mode, anyway? What's the developer story here?

The BIG issue is that you cannot disable entirely DSi mode. So you HAVE to make code specifically for it. Something that regular DS retail software doesn't need to do. Mapping 0x8000000 would remove the need for extra code.

A great solution would be to have an option (a make flag, maybe?) to disable it.

The actual issue I have is that it would lead to annoying checks for DSi mode on GBA-based code. For each and every function that accesses either the ROM or the SRAM. Luckily, there are only two of them that would require this in my case, but it could get annoying really quickly for bigger projects.

AntonioND:
Would it actually be that annoying? In all cases you need to check if there is an actual game or not anyway. You need a flag to stop reading from that slot if you have already seen that there is no game inserted. Adding the extra check is one function call at the point of checking if there is an actual game inserted, the rest of the code doesn't require changes.

If a DSi is detected, you tell the user "This is a DSi, GBA cartridge features are not supported" and set the "game detected" flag to false, and then you check that flag whenever you want to access the slot, which is something you're probably already doing.

asiekierka:
>The BIG issue is that you cannot disable entirely DSi mode. So you HAVE to make code specifically for it.

Not all DSi execution contexts allow to go back to DS mode. Pretty much anything that doesn't take over the firmware like Unlaunch does won't let you do so. On a CycloDS iEvolution, for example, you can't switch from DSi mode to DS mode once you've booted to the cartridge; you can only boot straight into DS mode. (The same went for savegame exploits of old.) 

>A great solution would be to have an option (a make flag, maybe?) to disable DSi mode

BlocksDS already supports this, if you want; add the `-h 0x200` argument to `ndstool` to make it generate a `.nds` that does not support DSi mode.

>The actual issue I have is that it would lead to annoying checks for DSi mode in GBA-based code. For each and every function that accesses either the ROM or the SRAM.

Having checks for whether a GBA cartridge is inserted is a good idea regardless, if you're accessing ROM or SRAM. It leads to faster and more reliable code for those people who aren't using a GBA cartridge. From there, it's trivial to extend that with a `if (!isDSiMode()) return false; /* never inserted */`.

Lorenzooone:
Okay. I'll put the check inside of a ifdef __NDS__ and just use that.

> `-h 0x200`
For anyone else reading this, setting this argument in ndstool will make it so your product won't work with some launchers, so beware.

PR #39, opened 2023-10-17
syscalls: remove ARM7-specific syscalls.c file

asiekierka:
* Move filesystem stubs to a separate `filesystem.c` file.
* Move `arm9/libc/sbrk.c` to `common/libc/sbrk.c`, incorporating ARM7-specific changes.
* Move the remaining differences into `common/libc/syscalls.c`.

Might be good to test some examples with this.


source/common/libc/sbrk.c
```diff
@@ -23,10 +23,14 @@ void *sbrk(int incr)
...
     if (heap_start == 0)
     {
+#ifdef ARM9
         if (fake_heap_start == NULL)
             heap_start = end;
         else
             heap_start = (uintptr_t)fake_heap_start;
+#else
+        heap_start = end;
+#endif
```
AntonioND, line 33: The resulting function in the case of the ARM7 doesn't match the previous version, are you sure this is right?

I get something a bit more like this, but the ifdefs make the code a bit hard to follow, I wonder if this is a function that we should leave as two functions like it was before.

```
void *sbrk(int incr)
{
    // Symbol defined by the linker
    extern char __end__[];
    const uintptr_t end = (uintptr_t)__end__;

    // Trick to get the current stack pointer
    register uintptr_t stack_ptr asm("sp");

    // Next address to be used. It is updated after every call to sbrk()
    static uintptr_t heap_start = 0;

#ifdef ARM9
    if (heap_start == 0)
    {
        if (fake_heap_start == NULL)
            heap_start = end;
        else
            heap_start = (uintptr_t)fake_heap_start;
    }

    // Current limit
    uintptr_t heap_end;

    if (fake_heap_end == NULL)
        heap_end = stack_ptr;
    else
        heap_end = (uintptr_t)fake_heap_end;
#endif

#ifdef ARM7
    if (heap_start == 0)
        heap_start = end;

    // Current limit
    uintptr_t heap_end = stack_ptr;
#endif

    // Try to allocate
    if (heap_start + incr > heap_end)
    {
        errno = ENOMEM;
        return (void *)-1;
    }

    uintptr_t prev_heap_start = heap_start;

    heap_start += incr;

    return (void *)prev_heap_start;
}
```

asiekierka: Should be fixed, now, in a similar but slightly different way.

AntonioND (approved):

PR #40, opened 2023-10-29
teak: Add libteak and ARM9 teak support

AntonioND:
libteak is a prototype library that can be used to develop binaries to be run in the DSP of the DSi. It requires ARM9 support, which is also added in this commit.

### DIFF
--- a/include/nds/system.h
+++ b/include/nds/system.h
@@ -457,11 +457,49 @@ void resetARM9(u32 address);
 
 // DSi SCFG registers
 
-#define REG_SCFG_ROM            *(vu16 *)0x4004000
-#define REG_SCFG_CLK            *(vu16 *)0x4004004
-#define REG_SCFG_RST            *(vu16 *)0x4004006
-#define REG_SCFG_EXT            *(vu32 *)0x4004008
-#define REG_SCFG_MC             *(vu16 *)0x4004010
+// SCFG_xxROM
+// ==========
+
+#define REG_SCFG_ROM            (*(vu16 *)0x4004000)
+
+#ifdef ARM7
+#define REG_SCFG_A9ROM          (*(vu8 *)0x4004000)
+#define REG_SCFG_A7ROM          (*(vu8 *)0x4004001)
+#endif // ARM7
+
+// SCFG_CLK
+// ========
+
+#define REG_SCFG_CLK            (*(vu16 *)0x4004004)
+
+#ifdef ARM9
+#define SCFG_CLK_ARM9_TWL       BIT(0)
+#define SCFG_CLK_DSP            BIT(1)
+#define SCFG_CLK_CAMERA_IF      BIT(2)
+#define SCFG_CLK_NWRAM          BIT(7) // Read only, set by ARM7
+#define SCFG_CLK_CAMERA_EXT     BIT(8)
+#endif // ARM9
+
+#ifdef ARM7
+#define SCFG_CLK_SDMMC          BIT(0)
+#define SCFG_CLK_NWRAM          BIT(7)
+#define SCFG_CLK_TOUCH          BIT(8)
+#endif // ARM7
+
+// SCFG_RST
+// ========
+
+#define REG_SCFG_RST            (*(vu16 *)0x4004006)
+
+#ifdef ARM9
+#define SCFG_RST_DSP_APPLY      (0 << 0)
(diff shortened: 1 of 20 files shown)


include/teak/types.h
```diff
@@ -0,0 +1,18 @@
...
+// Copyright (c) 2023 Antonio Niño Díaz
+
+#ifndef LIBTEAK_TYPES_H__
+#define LIBTEAK_TYPES_H__
+
+#include <stdint.h>
+
+#define BIT(n) (1 << (n))
+
+typedef volatile uint16_t vu16;
```
asiekierka, line 12: If we're going to add the common shorthands for types here, we might as well add all the `u/s/vu/vs` types from `ndstypes.h`.

include/nds/system.h
```diff
@@ -457,11 +457,47 @@ void resetARM9(u32 address);
...
-#define REG_SCFG_ROM            *(vu16 *)0x4004000
-#define REG_SCFG_CLK            *(vu16 *)0x4004004
-#define REG_SCFG_RST            *(vu16 *)0x4004006
-#define REG_SCFG_EXT            *(vu32 *)0x4004008
-#define REG_SCFG_MC             *(vu16 *)0x4004010
+#define REG_SCFG_ROM            (*(vu16 *)0x4004000)
+
+#define REG_SCFG_CLK            (*(vu16 *)0x4004004)
+
+#define SCFG_CLK_ARM9_134       (1 << 0)
```
asiekierka, line 464: These defines are largely duplicates.

See:

```c
#ifdef ARM9
#define SCFG_CLK_ARM9_TWL       BIT(0)
#define SCFG_CLK_DSP            BIT(1)
#define SCFG_CLK_CAMERA_IF      BIT(2)
#define SCFG_CLK_MBK_RAM        BIT(7)
#define SCFG_CLK_CAMERA_EXT     BIT(8)
// ...
```

include/nds/arm9/teak/twlwram.h
```diff
@@ -0,0 +1,95 @@
+// SPDX-License-Identifier: Zlib
+//
+// Copyright (C) 2020 Gericom
+
+#ifndef LIBNDS_NDS_ARM9_TEAK_TWLWRAM_H__
```
asiekierka, line 5: The nocash-proposed term here is `New Shared WRAM`, or `nwram` for short. I would standardize on that.

Also, it is not a Teak-specific functionality. I would move this out of `arm9/teak`.

*Also* (sorry for being pedantic), the NWRAM registers also exist on ARM7. WRAM A/B/C allocation is read-only, but address mapping (MBK6-8) are read-write and separate on ARM7.

It is also important to note that some homebrew launching methods do not support changing SCFG flags and may not support, for example, NWRAM, as a result.

Essentially, the changes I would do here are as follows:

* define a `common/nwram.h` and rename all occurences of `TWR_WRAM`, `TWLWRAM`, etc. to `NWRAM`
* Provide all defines and `nwramGetBlockAddress` and `nwramSetBlockMapping` (naming more consistent with libnds convention) on both ARM9 and ARM7,
* Provide `nwramMapSlotA/SlotB/SlotC` on ARM9 only.

include/teak/types.h
AntonioND: Done.

include/nds/system.h
AntonioND: Ah, I didn't notice it, too many definitions to track across different files. I've reorganized them a bit so that they are closer to their registers, and removed the duplicates.

include/nds/arm9/teak/twlwram.h
AntonioND: I've done everything except for the renaming of the MBK registers. I'm not sure about whether that's a good idea or not, given that GBATEK is the only reference we have, and it uses said names. This could be done in the future, anyway.

AntonioND:
I've also split the previous version of the PR into multiple commits so that they are easier to understand.

include/nds/arm9/teak/twlwram.h
asiekierka: At the very least, I would keep both around as synonyms. It makes code more readable.

AntonioND: I've created this issue: https://github.com/blocksds/sdk/issues/92 I wouldn't want to do it as part of this PR, it is quite complex already.

PR #41, opened 2023-11-05
Initial NitroFS support

asiekierka:
This uses the top 4 bits of the file descriptor as a "tag"; `0 << 28` means a FAT file, `1 << 28` means a NitroFS file.

Beyond that, everything else is handled by a pile of janky (but at least partially tested!) code written by myself.

I'm not perfectly happy with it, but it does fix the most important issue of NitroFAT: the low performance caused by the double indirection.

This also fixes bugs in `readdir()` and `rewinddir()` pertaining to the "beginning" and "end" of a directory being the same thing.

You may wish to move `cardReadUnaligned` into its own function in `card.c`, if you prefer.

Resolves https://github.com/blocksds/sdk/issues/66 , https://github.com/blocksds/sdk/issues/81 , https://github.com/blocksds/sdk/issues/88 .


AntonioND:
I've run a few tests with this example: https://github.com/blocksds/sdk/tree/11b6b357e0be59c154966dc6127dd4ab13437860/examples/filesystem/nitrofat_and_fat

My results are that:
- DSi: NitroFS works, as well as SD access. Loaded with nds-hb-menu.
- DS (with an Acekard 2i): It looks like the initial card read command works and the FAT table offsets are read correctly, but it fails to read the filesystem when I try to chdir() to the test folder.
- no$gba, melonDS, desmume: It looks like the unaligned card read function isn't reading anything, it returns zeroes, even though that the CPU that owns the bus is the right one for the cardRead() functions that are used internally.

asiekierka:
The emulator case should be fixed, now; in addition, the code has been made more compliant with Slot-1 protocol specifications. Supporting flashcarts in non-DLDI mode is out of scope.

AntonioND:
I've tested this again, on all my flashcards. It works on DSi, DeSmuMe, melonDS and no$gba.

However, out of my slot-1 cards, only my R4 (with wood R4) has worked. My AK2i, DSTWO and my clone AK2i can all see the FAT filesystem, but they can't see the nitroFS filesystem.

I'm going to take a look at it, but I'm not sure if I will have time to find the issue tonight.

AntonioND:
Every device I own works now.

AntonioND (approved):

PR #42, opened 2023-11-05
bios: inline BIOS SWI calls

asiekierka:
This uses a technique introduced by @felixjones in gba-hpp to inline BIOS SWI calls.

These specific ones aren't used *too* commonly in practice, but hey - a win's a win. However, testing with programs that make use of them (such as LZ/Huffman compression, math routines, etc) are required before merging.

### DIFF
--- a/include/nds/bios.h
+++ b/include/nds/bios.h
@@ -3,6 +3,7 @@
 // Copyright (C) 2005 Michael Noland (joat)
 // Copyright (C) 2005 Jason Rogers (dovoto)
 // Copyright (C) 2005 Dave Murphy (WinterMute)
+// Copyright (C) 2022-2023 gba-hpp contributors
 
 // BIOS functions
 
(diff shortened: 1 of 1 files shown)


AntonioND:
You also need to remove the ASM version from [here](https://github.com/blocksds/libnds/blob/6d2a95c4fd1b016cbdfa4a491426bbc037f747f3/source/common/biosCalls.s), right? My main worry is that if someone was using the BIOS functions from ASM code, this PR will break their code.

OR is your plan to leave both versions in and trust the compiler to always inline the function calls? If so we would need the ``alwaysinline`` attribute in case someone builds the code in a way that makes the compiler generate a standalone version of the static inline functions.

asiekierka:
My plan is to leave both versions for cases where `libnds` is used for languages other than C and C++ - that is, deliberately leaving symbol-level compatibility.

AntonioND:
That sounds good. As you mention in the original comment, I'll write a test for some SWIs before merging the PR, just in case.

AntonioND:
This should be enough to ensure that the PR hasn't broken anything: https://github.com/blocksds/sdk/commit/beeafca1122b42fb5eac3d2e8607f1162fddb394

Also, the keyboard of libnds is decompressed using SWI functions, so it's yet another test.

PR #45, opened 2023-11-06
fatfs: fix stat() on root directory not showing it as a directory

asiekierka:


AntonioND:
I assume this is not in the official fatfs, right?

PR #46, opened 2023-11-07
nitrofs: add Slot-2 support

asiekierka:
https://github.com/blocksds/sdk/issues/94

### DIFF
--- a/source/arm9/libc/nitrofs.c
+++ b/source/arm9/libc/nitrofs.c
@@ -54,9 +54,15 @@ static ssize_t nitrofs_read_internal(void *ptr, size_t offset, size_t len)
         fseek(nitrofs_local.file, offset, SEEK_SET);
         return fread(ptr, 1, len, nitrofs_local.file);
     }
+    else if (nitrofs_local.use_slot2)
+    {
+        sysSetCartOwner(BUS_OWNER_ARM9);
+        memcpy(ptr, (void*) (0x08000000 + offset), len);
+        return len;
+    }
     else
     {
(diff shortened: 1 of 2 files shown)


AntonioND:
Hmmmm I can't get DLDI to work at all on my SCSD. Well, no, I can't even get nds-hb-menu to run on my SCSD. I need to load the test ROM through nds-hb-menu because the default loader of the SCSD won't setup ``argv``. But that gives me white screens, like loading the test ROM directly. <https://github.com/blocksds/sdk/tree/0039642fb0009ad0ca975f1b5a9e77ec8ce46005/examples/filesystem/nitrofs_and_fat>

AntonioND:
Well, after a bit more testing I've seen that this actually works. I'll document how to get slot-2 devices working today, but this PR can be merged now.

AntonioND:
For reference, I've loaded that example ROM with nds-hb-menu. I took ``BOOT.NDS`` from the ``hbmenu.zip`` archive, patched it with ``scsd_moon.dldi``, and then concatenated ``ndsmall.bin`` with ``BOOT.NDS``: ``cat ndsmall.bin BOOT.NDS > hbmenu.sc.nds``. You can find ``ndsmall.bin`` in Darkfader's website. With this patched nds-hb-menu I can load the unpatched example ROM and have access to nitroFS and FAT.

Also, concatenating ``ndsmall.bin`` with the example ROM can get it to boot with nitroFS support. This is the same effect as generating the NDS file adding ``-h 0x200`` to the invocation of ``ndstool``.

PR #47, opened 2023-11-07
fatfs, nitrofs: properly populate st_dev and st_ino

asiekierka:
As per [POSIX specifications](https://pubs.opengroup.org/onlinepubs/009695299/basedefs/sys/stat.h.html), 

>The st_ino and st_dev fields taken together uniquely identify the file within the system.

This PR implements that for FAT and NitroFS filesystems as such:

* For FAT filesystems, `st_dev` is 0 (DLDI) or 1 (DSi SD), while `st_ino` is the starting cluster of the file;
* For NitroFS filesystems, `st_dev` is 2, while `st_ino` is the unique ID of the file within the `.nds` file.

I believe it to be unwise to document the exact meaning of these values outside of code, as users should not rely on them.

Whether this should be merged or not, I'm not too sure; it *does* resolve https://github.com/blocksds/sdk/issues/85 , but I don't think code should be relying on this... But here it is, anyway.

### DIFF
--- a/source/arm9/libc/fatfs/ff.c
+++ b/source/arm9/libc/fatfs/ff.c
@@ -4799,6 +4799,11 @@ FRESULT f_stat (
 			/* BlocksDS - support origin directory stat(), write device/cluster information */
 			if (fno) {
 				fno->fpdrv = dj.obj.fs->pdrv;
+#if FF_FS_EXFAT
+				if (fs->fs_type == FS_EXFAT) {
+					fno->fclust = 0; /* FIXME */
+				} else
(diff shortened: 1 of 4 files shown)


AntonioND:
> Whether this should be merged or not, I'm not too sure; it does resolve https://github.com/blocksds/sdk/issues/85 , but I don't think code should be relying on this... But here it is, anyway.

Yes, I agree this is quite hacky, but the tools that use this field are quite hacky by design. After all, we're talking about the DS: it's super limited.

PR #49, opened 2023-11-08
sound, timers: round frequencies to nearest, fix documentation

asiekierka:
… frequency

This is an attempt at resolving https://github.com/blocksds/sdk/issues/42 - of course, some testing would be required. `libxm7` works fine, so the `SOUND_FREQ` values are sound - but this also impacts `TIMER_FREQ`, and possibly code size/performance. Would be good to look for some homebrew that make heavy use of those.

~~Two changes are made, currently:

* ~~Synchronize `TIMER_FREQ` and `SOUND_FREQ` definitions with each other, so that the same frequency leads to the same speed in both cases.~~
* Make both defintions round to the nearest frequency, instead of rounding up.

Some things to consider:

* Is rounding to nearest, that is *including rounding down*, for timers acceptable in the first place? It's good for sampled audio, but what about other usecases?
* ~~`TIMER_FREQ` has 2x the precision of `SOUND_FREQ`; combined with the above question, I'd consider adding a `TIMER_SOUND_FREQ` which ensures that the speed of `TIMER_SOUND_FREQ(n)` == `SOUND_FREQ(n)` for any `n`.~~ Not in this PR.

### DIFF
--- a/include/nds/arm7/audio.h
+++ b/include/nds/arm7/audio.h
@@ -28,7 +28,7 @@ extern "C" {
 #include <nds/timers.h>
 
 #define SOUND_VOL(n)        (n)
-#define SOUND_FREQ(n)       TIMER_FREQ_SHIFT(n, 1)
+#define SOUND_FREQ(n)       TIMER_FREQ_SHIFT(n, 1, 1)
 #define SOUND_ENABLE        BIT(15)
 
 #define SOUND_REPEAT        BIT(27)
--- a/include/nds/arm7/audio.h
+++ b/include/nds/arm7/audio.h
@@ -25,9 +25,10 @@ extern "C" {
 
 #include <nds/arm7/serial.h>
 #include <nds/system.h>
+#include <nds/timers.h>
 
 #define SOUND_VOL(n)        (n)
-#define SOUND_FREQ(n)       ((-0x1000000 / (n)))
+#define SOUND_FREQ(n)       TIMER_FREQ_SHIFT(n, 1)
 #define SOUND_ENABLE        BIT(15)
 
 #define SOUND_REPEAT        BIT(27)
--- a/include/nds/timers.h
+++ b/include/nds/timers.h
@@ -117,7 +117,7 @@ typedef enum {
 /// - ClockDivider_256  = 8
 /// - ClockDivider_1024 = 10
 /// - ARM7 sound timer  = 1
-#define TIMER_FREQ_SHIFT(n, shift) (-((BUS_CLOCK >> (shift)) + (((n - 1)) >> 1)) / (n))
+#define TIMER_FREQ_SHIFT(n, divisor, shift) ((-((BUS_CLOCK >> (shift)) * (divisor)) - ((((n) + 1)) >> 1)) / (n))
 
 /// A macro that calculates TIMER_DATA(n) settings for a given frequency of n.
 ///
@@ -134,7 +134,7 @@ typedef enum {
 /// Min frequency is: ~511 Hz
 ///
 /// @note Use the appropriate macro depending on the used clock divider.
-#define TIMER_FREQ(n)    TIMER_FREQ_SHIFT(n, 0)
+#define TIMER_FREQ(n)    TIMER_FREQ_SHIFT(n, 1, 0)
 
 /// A macro that calculates TIMER_DATA(n) settings for a given frequency of n.
 ///
@@ -151,7 +151,7 @@ typedef enum {
 /// Min frequency is: 8 Hz
 ///
 /// @note Use the appropriate macro depending on the used clock divider.
-#define TIMER_FREQ_64(n) TIMER_FREQ_SHIFT(n, 6)
+#define TIMER_FREQ_64(n) TIMER_FREQ_SHIFT(n, 1, 6)
 
 /// A macro that calculates TIMER_DATA(n) settings for a given frequency of n.
 ///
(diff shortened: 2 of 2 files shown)


Lorenzooone:
I just had to struggle with it here: https://github.com/blocksds/maxmod/blob/57953384ec570a8363542d5c89d575095e874bbe/source_ds7/mm_mixer_super.c#L40

Keeping the values synchronized requires accounting for the various rounding errors, as well as how many samples are there to consume for the sound system. Something that doesn't seem to be happening?! I may be reading this wrong, though.

(I also don't like how you block the usage of sub 256hz for sound frequencies. Maybe offer some kind of alternative that does the missing range using the other types of timers?)

Lorenzooone:
While we're at it, another issue with all of these macros is that they only allow integer frequencies.
Something like 194.8125 Hz is not possible.
You can check here how I ended up doing it: https://github.com/blocksds/maxmod/blob/57953384ec570a8363542d5c89d575095e874bbe/source_ds7/mm_mixer_super.c#L34

asiekierka:
> (I also don't like how this blocks the usage of sub 512hz for sound frequencies, despite how rare their usage may be. Maybe offer some kind of alternative that does the missing range using the other types of timers?)

The hardware blocks it, though? The timer frequency value can only go up to 65535, which is equal to `(33513982 / (2 * 65535))` Hz, or 255.7 Hz. You can't go lower than that for sampled audio.

asiekierka:
I've added a tweak to allow pre-multiplying BUS_CLOCK to act as a specific divisor for a fixed-point value; though for something like `TIMER_FREQ`, due to 32-bit register limitations, it will only work for divisors <= `64`.

I've also fixed up the algorithm a bit.

asiekierka:
`TIMER_SOUND_FREQ` needs to be rethought: it's a common usecase to, say, be playing something at `24000` Hz, but only wanting to synchronize every `24000 / 256` Hz.

asiekierka:
I removed TIMER_FREQ_SOUND as I'm iffy on it - the rest seems solid to me, and there's no point in having the PR sit for this many months.

AntonioND:
Hmmm... I've taken a look at the PR and it's pretty clear that we need tests that use the timers. Right now there is only one example that uses the tests, but it clearly doesn't cover many cases...

include/nds/arm7/audio.h
```diff
@@ -25,9 +25,10 @@ extern "C" {
 
 #include <nds/arm7/serial.h>
 #include <nds/system.h>
+#include <nds/timers.h>
 
 #define SOUND_VOL(n)        (n)
-#define SOUND_FREQ(n)       ((-0x1000000 / (n)))
+#define SOUND_FREQ(n)       TIMER_FREQ_SHIFT(n, 1, 1)
```
AntonioND, line 31: This is probably being a bit pedantic, but can you wrap that `n` in parentheses?

include/nds/timers.h
```diff
@@ -108,22 +108,33 @@ typedef enum {
...
 /// ```
 ///
-/// Max frequency is: 33554432 Hz
-/// Min frequency is: 512 Hz
+/// Max frequency is: 33513982 Hz
+/// Min frequency is: ~511 Hz
 ///
 /// @note Use the appropriate macro depending on the used clock divider.
-#define TIMER_FREQ(n)    (-BUS_CLOCK / (n))
+#define TIMER_FREQ(n)    TIMER_FREQ_SHIFT(n, 1, 0)
```
AntonioND, line 137: Same here.

include/nds/timers.h
```diff
@@ -132,15 +143,15 @@ typedef enum {
...
+/// timerStart(0, ClockDivider_64, TIMER_FREQ_64(500), timerCallBack);
 /// ```
 ///
-/// Max frequency is: 524288 Hz
+/// Max frequency is: ~523656 Hz
 /// Min frequency is: 8 Hz
 ///
 /// @note Use the appropriate macro depending on the used clock divider.
-#define TIMER_FREQ_64(n) (-(BUS_CLOCK >> 6) / (n))
+#define TIMER_FREQ_64(n) TIMER_FREQ_SHIFT(n, 1, 6)
```
AntonioND, line 154: And here.

include/nds/timers.h
```diff
@@ -149,15 +160,15 @@ typedef enum {
...
 /// ```
 ///
-/// Max frequency is: 131072 Hz
-/// Min frequency is: 2 Hz
+/// Max frequency is: ~130914 Hz
+/// Min frequency is: ~2 Hz
 ///
 /// @note Use the appropriate macro depending on the used clock divider.
-#define TIMER_FREQ_256(n) (-(BUS_CLOCK >> 8) / (n))
+#define TIMER_FREQ_256(n) TIMER_FREQ_SHIFT(n, 1, 8)
```
AntonioND, line 171: And here.

include/nds/timers.h
```diff
@@ -170,11 +181,11 @@ typedef enum {
...
 /// ```
 ///
-/// Max frequency is: 32768 Hz
-/// Min frequency is: 0.5 Hz
+/// Max frequency is: ~32728 Hz
+/// Min frequency is: ~0.5 Hz
 ///
 /// @note Use the appropriate macro depending on the used clock divider.
-#define TIMER_FREQ_1024(n) (-(BUS_CLOCK >> 10) / (n))
+#define TIMER_FREQ_1024(n) TIMER_FREQ_SHIFT(n, 1, 10)
```
AntonioND, line 188: And here.

AntonioND:
I've improved the timers example here: https://github.com/blocksds/sdk/commit/d082cc8c3e489b8f46a137cef6b6ab4f3b468826

After that test, I'm happy with the PR.

When you take care of the parentheses, can you reduce the number of commits in this branch? We could have one commit with the changes to the timers and another one for the changes to the sound (or one commit with both changes, I don't care). There is no point in having a few commits that add things and others that remove the same things in the same PR, and I don't like squashing branches so that the author of the commits remains the original author (if I squash the branch, GitHub will become one of the authors).

AntonioND:
Actually, forget about the comments of the parentheses...

AntonioND:
I've squashed the branch into a single commit so that there is less noise: https://github.com/blocksds/libnds/commit/fe84445c525a14adfd9a8a34ff0ae4925c6a2a37

PR #50, opened 2023-11-09
nitrofs: add missing nitrofs_initialized guards

asiekierka:


AntonioND:
Ah, right, that should make everything safer, thanks.

PR #52, opened 2023-11-10
firmware: add readJEDEC()

lifehackerhansol:
This is a helper function to read the flash JEDEC values.

A use case for this function is to retrieve the size of the flash; this can be useful in verifying internal hardware, or verify the size of the flash before doing a read of the entire device.

### DIFF
--- a/source/arm7/firmware.c
+++ b/source/arm7/firmware.c
@@ -39,6 +39,24 @@ void readFirmware(u32 address, void *destination, u32 size)
     leaveCriticalSection(oldIME);
 }
 
+int readJEDEC(u8 *destination, u32 size) {
+    if (destination == NULL)
+        return -1;
+
+    // JEDEC is always 3 bytes.
+    if (size < 3)
+        return -2;
+
+    int oldIME = enterCriticalSection();
+    REG_SPICNT = SPI_ENABLE | SPI_BYTE_MODE | SPI_CONTINUOUS | SPI_DEVICE_FIRMWARE;
+    readwriteSPI(FIRMWARE_RDID); // get JEDEC
+    for (int i = 0; i < 3; i++) {
+        destination[i] = readwriteSPI(0);
+    }
+    REG_SPICNT = 0;
+    leaveCriticalSection(oldIME);
+}
+
 static int writeFirmwarePage(u32 address, u8 *buffer)
 {
     u8 pagebuffer[256];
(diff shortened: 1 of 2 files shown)


lifehackerhansol:
PR isn't tested yet; this is a copy of https://github.com/DS-Homebrew/dsbf_dump/blob/c9d82b1bbddef4194feac5bcd0e20f9ba52d7212/arm7/source/main.c#L20C1-L34C2 that made sense to upstream at least to me.

AntonioND:
I assume all we need to test this is to set some japanese characters as user name and see if the read string is correct, right?

lifehackerhansol:
This only returns the JEDEC values as seen in https://problemkaputt.de/gbatek-ds-firmware-serial-flash-memory.htm. Doesn't have much to do with actual flash contents.

There isn't much to test with this function other than checking whether the returned values match the ones seen on GBATEK (or maybe you have a DS with an unknown flash chip?)

AntonioND:
Ah, sorry, my brain thought about JIS when I read your PR and thought this was about Japanese characters. 😅

source/arm7/firmware.c
```diff
@@ -39,6 +39,21 @@ void readFirmware(u32 address, void *destination, u32 size)
     leaveCriticalSection(oldIME);
 }
 
+void readJEDEC(u8 *destination, u32 size) {
```
AntonioND, line 42: This function should return an ``int``: 0 on success, -1 (for example) on error. The other functions in this file should do that too, but let's ignore that for the time being.

Also, you should check if ``destination`` is NULL.

asiekierka:
More broadly speaking, I'd like to see the code for reading firmware refactored entirely, to at least partially unify it with code for reading and/or writing to cartridge flash/EEPROM data, if possible.

It's the same protocol; we should be able to control for this via a flag byte.

source/arm7/firmware.c
lifehackerhansol: I kind of arbitrarily added error codes for destination and size. (Call me incompetent, but never actually implemented error codes in an expected way before.)

AntonioND (approved):

AntonioND:
There was a small warning in the function but I've fixed it here: https://github.com/blocksds/libnds/commit/a9c99219c954fecfee3c7f8c9158e39c20ac64a9

PR #54, opened 2023-11-11
userSettings: minor readFirmware() optimization

asiekierka:
Two fewer SPI bus reinitializations.


AntonioND:
You need to fix the declaration in the header or this will fail to build.

asiekierka:
Fixed!

AntonioND:
Thanks!

PR #55, opened 2023-11-11
nitrofs: fix file traversal not working without running a chdir() first

asiekierka:
Let us never speak of this again.


AntonioND:
Thanks!

PR #56, opened 2023-11-12
nitrofs: fix missing fclose() if pointed to NitroFS-less .nds file

asiekierka:
This is unlikely to be hit in production, and so doesn't warrant an 0.10.3.


AntonioND:
Thanks!

PR #58, opened 2023-11-13
DLDI/cache improvements.

asiekierka:
* Patched FatFS to only cache FAT/directory reads, never file ones.
* Adjusted DLDI stub to changes in https://github.com/blocksds/sdk/pull/98
* Use unused DLDI driver area as a sector cache by default. On DSi/melonDS, this means 31 sectors; on DesMuME with the tiny R4 driver, this means 30 sectors; on real hardware, this will probably mean something akin to 16-24 sectors, depending on driver. It remains to be seen if these amounts are sufficient; however, explicitly requesting a larger sector cache is still supported. This saves ~20KB of dynamically allocated RAM for software using FatFS.


asiekierka:
This fixes https://github.com/blocksds/sdk/issues/89

source/arm9/libc/fatfs/cache.c
```diff
@@ -22,33 +23,48 @@ static cache_entry_t *cache_entries;
...
-    cache_mem = malloc(num_sectors * FF_MAX_SS);
-    if (cache_mem == NULL)
-    {
-        free(cache_entries);
-        return -1;
-    }
+        // If sufficient, use the space at the end of the DLDI stub
+        cache_mem = num_sectors <= stub_space_sectors
+            ? dldiGetStubEnd() - (num_sectors << 9)
+            : malloc(num_sectors * FF_MAX_SS);
```
AntonioND, line 57: Maybe add a comment here saying that the space allocated by malloc() will never be freed, so this can't cause memory leaks.

source/arm9/libc/fatfs/cache.c
```diff
@@ -22,33 +23,48 @@ static cache_entry_t *cache_entries;
...
-    if (cache_mem != NULL)
+    if (cache_mem != NULL && (uintptr_t)cache_mem > (uintptr_t)dldiGetStubEnd())
         free(cache_mem);
 
-    cache_entries = calloc(num_sectors, sizeof(cache_entry_t));
-    if (cache_entries == NULL)
-        return -1;
+    int32_t stub_space_sectors = (dldiGetStubEnd() - dldiGetStubDataEnd()) >> 9;
+    if (num_sectors < 0) {
+        num_sectors = stub_space_sectors;
```
AntonioND, line 42: Hmmm is this a good idea? You're allocating less space for the cache of bigger DLDI drivers.

I think we should do this in a different way. Instead of having a big RAM chunk that we use as cache, we can have an array of pointers to sectors. Then, we allocate as many sectors as we can in the space of the DLDI driver, and the rest in the heap.

AntonioND: Well, we don't need an array of pointers, maybe just two pointers: One to the free space in the DLDI driver and another one to the extra space in the heap.

asiekierka: We can get away with just two pointers, yes.

>Hmmm is this a good idea? You're allocating less space for the cache of bigger DLDI drivers.

This is deliberate.

a) The vast majority of currently used DLDI drivers are <=8KB, which leaves a cache of 16-24 sectors.
b) On DSi, the DLDI driver is populated by a stub, which leaves a cache of 31 sectors.

The previous default cache size was 40 sectors; as the cache has been optimized to only be invoked for FAT directory/cluster table reads, I don't think it needs to be as large to get the same performance. Unless there's a significant performance gain from a larger cache, I think it is fine to make this approach the default (as using the DLDI reserved space for a sector cache is always "free", RAM-wise). Nevertheless, allocating larger caches is always possible - be it by defining your own `__dldi_size` symbol to reserve a 32KB area, or by passing a larger size to fatInit().

PR #62, opened 2023-11-18
slot2: New peripheral detection logic, including external RAM support…

asiekierka:
… and more rumble support

`rumble.c` has been rewritten from the ground up to use the new detection logic; piano/paddle/guitarGrip have been largely left alone.

More testing required.


asiekierka:
Some amount of testing has been received. It's not ideal, but it is a start.

asiekierka:
Fixes https://github.com/blocksds/sdk/issues/99 and https://github.com/blocksds/sdk/issues/100

AntonioND:
Thanks! I'm sure that when more people are using the code they will be more likely to find bugs than if we leave this in a branch without merging it.

PR #65, opened 2023-11-19
mpu: Add helpers to enable/disable data cache for slot-2 memory

AntonioND:


asiekierka:
Aside from the discourse on write-back, LGTM.

include/nds/system.h
```diff
@@ -450,6 +450,14 @@ void *memCached(void *address);
 /// @return a pointer to the uncached mirror of that address.
 void *memUncached(void *address);
 
+/// Enables data cache for the DS slot-2 memory region.
```
asiekierka, line 453:
```
/// Enables data cache for the DS slot-2 memory region.
///
/// Note that this is not safe to enable if you're using Slot-2 memory for
/// purposes other than external RAM, such as rumble or other I/O.
///
/// @param write_back Set as write-back. This allows writes to skip
/// the slow Slot-2 bus, at the expense of requiring a full memory flush
/// when calling peripheralSlot2DisableCache().
```

include/nds/system.h
```diff
@@ -450,6 +450,14 @@ void *memCached(void *address);
...
+///
+/// @param write_back Set as write-back.
+void peripheralSlot2EnableCache(bool write_back);
+
+/// Disable data cache for the DS slot-2 memory region.
```
asiekierka, line 458:
```
/// Disable data cache for the DS slot-2 memory region.
/// If write-back was enabled, additionally clear and flush the data cache.
///
/// Note that flushing the data cache may still be required to remove stale
/// "read" pages, such as if you write to external RAM uncached and then
/// re-enable cache.
```

AntonioND: Done

include/nds/system.h
AntonioND: Done

PR #66, opened 2023-11-19
camera: Wrap functions to prevent crash in NDS mode

asiekierka:
No worries, does not justify a new release - but it does match existing libnds behaviour to check for this.


AntonioND:
Yeah, this is a good idea.

PR #73, opened 2023-12-06
videoGL: fix header warning in stricter build modes

asiekierka:


AntonioND:
So I really don't like the fact that we're fixing warnings for repositories that enable random warnings for severely outdated compilers. I think that a better fix in this case is to move ``glCallList()`` to ``videoGL.c``, the function is too big to make sense as an "always inline" function. Do you mind doing that change in this PR?

asiekierka:
Done. I've also found a major size/speed boost in the implementation of `glMaterialShinyness()`. Are there any other functions you think should go into the C file? You've worked more with the 3D engine than I have.

AntonioND:
> As discussed with the project lead, it was enabled with the need to maintain support for MSVC, which only gained C99/C11/etc. support in 2017 (!). I wouldn't call them "severely outdated" just because they're not up to date on standards; GCC and Clang still don't have full C++20 support, for example, due to the whole "modules" thing.

C99 is 23 years old, and ``-Wdeclaration-after-statement`` is only needed to keep compatibility with C90 because C99 doesn't need that kind of syntax. MSVC 2017 is 6 years old itself, so it took them 18 years to add a super simple feature (because they decided to not care about C and focus on C++, which is fair). If modern versions of MSVC support C99, then using MSVC 2017 is choosing to stick to C90, which is 33 years old. Not even Microsoft supports that version of MSVC, and support for MSVC 2019 was dropped in 2022.

I've read the [MSVC readme of megazeux](https://github.com/AliceLR/megazeux/blob/e62e3b84c01da151d8061480e3742b6f29a99700/arch/msvc/README.txt), the project that you mentioned in https://github.com/blocksds/sdk/issues/4#issuecomment-1843280724, and I think that they can achieve the same cross-compiler compatibility goals even if they update to a more modern MSVC.

However, if they want to stick to MSVC 2017, I agree that enforcing C90 in common code is a reasonable choice. What I don't think it's reasonable is to enforce that in platform code for a platform that can build with modern GCC.

In general, I think it's extremely unfair to expect everyone to stick to C90 because you want to support a very old version of MSVC. While I generally like to have as much compatibility as possible (specially with devkitARM projects) I think that it's important to take advantage of new versions of compilers and languages. This is the reason why I'm complaining.

What's the reason to stick to MSVC 2017, anyway? To be able to build to old versions of Windows or something?

> I would object more perhaps if the fix wasn't trivial (we already take larger measures to preserve compatibility, what with `static_assert`), but it is trivial and doesn't hurt us; certainly doesn't hurt us any more than downstream projects having to add exceptions just to deal with BlocksDS-specific quirks (no other platform in the similar league has this issue: not 3DS, not PSP, not Wii, not Wii U, not Switch, not Vita, etc).
> 
> Regardless, I agree that the better fix is to make fewer functions "always inline" - particularly functions which don't benefit significantly from inlining optimizations or functions which can be called much more than once (`glCallList()` is among them).

I haven't seen any C project that uses C90 syntax on purpose for years. I guess that the libraries of the platforms you mention don't have any significant code in header files and that's why they don't trigger warnings. That's why the fix I'm willing to accept is to move code away from headers and into C files (and potentially let LTO optimize it).

PR #75, opened 2023-12-16
filesystem: general improvements

asiekierka:
* Dot and dot-dot entries are now properly emitted in readdir(). In line with common software expectations, they mirror standard FAT filesystem behaviour, that is are present for all subdirectories. For NitroFS, these entries are emulated accordingly.
* The "d_ino" field in readdir() output is now correctly populated, to match stat() and fstat().
* New functions: `nitroFSOpenById()`, `nitroFSFopenById()`. These allow opening a file by its FAT table ID directly, skipping the cost of any required directory lookups.
* Allow opening NitroFS filesystems which only have a FAT table, but no FNT table, using the above API.

I wasn't sure whether to consider https://github.com/blocksds/sdk/issues/69 a bug or not, but @LunarLambda's analysis makes it clear that despite POSIX saying otherwise, the common Linux-targetting software expectation is for these entries to be present.

### DIFF
--- a/include/filesystem.h
+++ b/include/filesystem.h
@@ -42,6 +42,33 @@ void nitroFSExit(void);
 /// @return 0 if the initialization was successful, a non-zero value on error.
 int nitroFSInitLookupCache(uint32_t max_buffer_size);
 
+/// Open a NitroFS file descriptor directly by its FAT offset ID.
+///
+/// This FAT offset ID can be sourced from functions like @see stat,
+/// @see fstat or @see readdir - it is equivalent to the st_ino/d_ino value.
+///
+/// In all other functions, this file descriptor behaves identically to one
+/// sourced from @see open , and should be closed likewise.
+///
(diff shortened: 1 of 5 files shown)


LunarLambda:
> [...] that is are present for all subdirectories

Worth noting that Linux emits .. entries even for the root directory (which always links to itself). On FAT that entry isn't physically present, but I don't currently remember what Windows does, whether it emits it or not.

asiekierka:
> Worth noting that Linux emits .. entries even for the root directory (which always links to itself).

I think that's a needless complication code-wise (as I would have to add additional dot-dot emulation code to FAT `readdir()`), while the current solution does fit the requirements of the POSIX standard as well as the typical use of dot-dot in `readdir()`.

It is a good note, but not behaviour software should really be reliant on.

LunarLambda:
Completely fair, just figured I'd mention it for completeness. I do think this works totally fine.

AntonioND:
Thanks for this!

PR #80, opened 2024-01-21
grf: Add helper to load GRF files

AntonioND:
This is one of the formats that GRIT can generate, and it's useful because it packs multiple graphics blobs as well as metadata such as compression type and size.


AntonioND:
The idea is that GRF is a  pretty reasonable format that we could standardize for NDS graphics.

AntonioND:
https://github.com/blocksds/libnds/commit/e0f026f6ee827341e6a5c4cf8f570f5aef05453e

PR #83, opened 2024-01-28
slot2tilt: Fix API to actually function.

asiekierka:


AntonioND:
I've just tested this with my Yoshi cartridge, it works.

PR #84, opened 2024-01-29
card: nitroFS: Slot-1 NitroFS reads can be done from ARM7

AntonioND:
This commit removed the ability of NitroFS Slot-1 reads to be done from the ARM7:

Commit 4514e731afff ("nitrofs: remove ARM7 card read support (only worked with main RAM), fix ARM9 card read support a lot").

However, DLDI or DSi SD reads could still be done from the ARM7.

The result is that games behave differently in emulators than in real hardware. The filesystem asynchronous loading example found in ``examples/filesystem/async_loading`` would work on real hardware, or when emulating DLDI/DSi SD, but not when emulating Slot-1 card reads.

This commit reintroduces that support so that users can optionally ask the ARM7 to do the reads. This isn't important just to save cycles in the ARM9, but to let multithreading work the same way no matter where NitroFS is reading files from (Slot-1 card commands, direct Slot-2 memory reads, DLDI or DSi SD).

### DIFF
--- a/include/filesystem.h
+++ b/include/filesystem.h
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Zlib
 //
-// Copyright (c) 2023 Antonio Niño Díaz
+// Copyright (c) 2023-2024 Antonio Niño Díaz
 
 #ifndef LIBNDS_FILESYSTEM_H__
 #define LIBNDS_FILESYSTEM_H__
@@ -69,6 +69,17 @@ int nitroFSOpenById(uint16_t id);
 ///
 /// @return A valid file pointer; NULL on error.
 FILE *nitroFSFopenById(uint16_t id, const char *mode);
+
+/// Select the CPU that will do the NitroFS reads.
+///
+/// When NitroFS detects it's running in an official cartridge, or in an
+/// emulator, this function lets you define the CPU that is in charge of reading
+/// from the cart. By default the ARM7 is in charge of reading the cart. This
+/// function lets you switch to using the ARM9 as well. You may switch between
+/// CPUs at runtime, but be careful to not switch while the card is being read.
+///
+/// @param use_arm9 Set to true to use the ARM9, false to use the ARM7.
+void nitroFSSetReaderCPU(bool use_arm9);
 
 #ifdef __cplusplus
 }
--- a/include/filesystem.h
+++ b/include/filesystem.h
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Zlib
 //
-// Copyright (c) 2023-2024 Antonio Niño Díaz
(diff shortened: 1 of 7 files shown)


asiekierka:
Ideally, I'd like to see a world in which such asynchronous reads are handled on ARM9 using DMA (or NDMA!) when available, as to not burden either CPU and avoid FIFO latency. This would, however, require a major breaking change to libnds (having a system-wide DMA and timer allocator) and is thus probably not advisable for the scope of BlocksDS as it stands.

> This isn't important just to save cycles in the ARM9, but to let multithreading work the same way no matter where NitroFS is reading files from (Slot-1 card commands, direct Slot-2 memory reads, DLDI or DSi SD).

People should not write code that depends on the particulars of internal multithreading behavior, as it can always change. What if the sector you're reading from happens to be cached? What if the DLDI driver is ARM9-only? Then multithreading won't work the same way already.

In particular, I am skeptical of the value of calling `memcpy()`, of all functions, on the ARM7 only to jump back to the ARM9, and the associated code cost of supporting an usecase more niche than "ARM9-only DLDI drivers".

AntonioND:
> Ideally, I'd like to see a world in which such asynchronous reads are handled on ARM9 using DMA (or NDMA!) when available, as to not burden either CPU and avoid FIFO latency. This would, however, require a major breaking change to libnds (having a system-wide DMA and timer allocator) and is thus probably not advisable for the scope of BlocksDS as it stands.

I wonder if we could just let users call a function to setup DMA card reads, and let them select a timer and DMA channel. They would be responsible for not using them for anything else, though.

However, this would only really help in emulators unless DLDI drivers start using that too.

> People should not write code that depends on the particulars of internal multithreading behavior, as it can always change. What if the sector you're reading from happens to be cached? What if the DLDI driver is ARM9-only? Then multithreading won't work the same way already. Any code which relies on this to function is just bad code; it should call `yield()` on its own if it really needs to.

Yeah, I think I agree with this, I'll modify the example to call ``yield()``.

> In particular, I am skeptical of the value of calling `memcpy()`, of all functions, on the ARM7 only to jump back to the ARM9, and the associated code cost of supporting an usecase more niche than "ARM9-only DLDI drivers".

That's also fair, I will remove that code.

> The Slot-1 card command ARM9->ARM7 jump is probably acceptable for now.

I think I'm going to keep it, because it looks like the DMA card read option is a bit unreachable for the time being.

 I'd love to be proven wrong, though. Maybe we could make some standard interface to allocate a DMA channel and timer for card reads, and then DLDI drivers and libnds card read functions can use them. However, we both know that it's unlikely many people will see the benefits from this.

AntonioND:
https://github.com/blocksds/libnds/commit/8365b837c9c2049134b5485b402d7722585a2681

PR #85, opened 2024-01-30
mpu: Correctly detect debugger DSi RAM

AntonioND:
Bits 14 and 15 of SCFG_EXT9 and SCFG_EXT7 are R/W, and the loader can set them to any value it wants. This only affects what is physically connected to the bus at addresses 0x0D000000-0x0DFFFFFF (and 0x0C000000-0x0CFFFFFF if we want to run in DS mode, but that's not relevant for this patch).

When SCFG_EXT9 specifies that there are 32 MB of RAM available, a retail DSi has open bus in 0x0D000000-0x0DFFFFFF, and a debugger DSi has additional RAM.

In order to actually detect how much RAM there is, it is required to actually test if it's there. It is required to set the RAM size field of SCFG_EXT9 to 32 MB, and try to write/read from the region where the additional 16 MB would be stored.

Once the size has been determined we can use the MPU to lock down the uncached RAM region (16 MB for retail, 32 MB for debugger).

This patch also modifies the way DSi mode is detected. We can't rely on the loader giving us a sensible value in SCFG_EXT9, so the code has switched to using SCFG_A9ROM to detect DSi mode, like the crt0 does.

The old code assumed that the value in SCFG_EXT9 would be correct for retail and debugger units, which wasn't the case. Loaders set the field to 32 MB, so libnds was always mapping 32 MB of RAM even in retail units.

### DIFF
--- a/source/arm9/system/mpu_setup.s
+++ b/source/arm9/system/mpu_setup.s
@@ -56,7 +56,7 @@
 // but it isn't possible to map everything at 0x02000000 because shared WRAM is
 // mapped at 0x03000000, so there are only 16 MB available. The last 16 MB of
 // the main RAM can only be accessed from their non-cachable mirror. This isn't
-// a problem in a regular consumer DSi because it has exactly 16 MB of RAM.
+// a problem in a regular retail DSi because it has exactly 16 MB of RAM.
 //
 // Legend:
 //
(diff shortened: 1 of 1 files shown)


asiekierka:
> We can't rely on the loader giving us a sensible value in SCFG_EXT9, so the code has switched to using SCFG_A9ROM to detect DSi mode, like the crt0 does.

Personally, I'd say we can rely on it giving us a value that indicates 16/32MB of RAM, as per the SCFG analysis I did in the relevant issue. Then we just need to narrow it down to 16 or 32MB of RAM.

AntonioND:
> Personally, I'd say we can rely on it giving us a value that indicates 16/32MB of RAM, as per the SCFG analysis I did in the relevant issue. Then we just need to narrow it down to 16 or 32MB of RAM, and going down from 32 to 16 if so.

True, but why trust the value at all if we can use SCFG_A9ROM like commercial games (and the crt0)?

Is there any case of DSi entrypoint that wouldn't let us access the additional RAM? The RAM size bits are always unlocked.

asiekierka:
That's a fair counterpoint.

PR #88, opened 2024-03-06
videoGL: Cast integer to GL_TEXTURE_SIZE_ENUM explicitly

nbvdkamp:
The line in question gave an invalid conversion error when included from C++. This change ensures it is valid C and C++.


AntonioND:
Thanks!

PR #89, opened 2024-03-19
libc: fix bug in readdir()

NovaSquirrel:
`readdir()` should set the index to `INDEX_END_OF_DIRECTORY` when it hits the end of a directory, not `INDEX_NO_ENTRY`. This incorrect index caused a behavior where calling `seekdir()` after hitting the end of the directory would then cause it to attempt to read past the end of the directory, fail, and act as if you had just called `rewinddir()` instead.


AntonioND:
Thanks!

PR #91, opened 2024-04-12
keyboard: Replace keyboardEnd() with keyboardExit() which also deinit…

asiekierka:
…ializes the pointer, make keyboardInit() not link in the default keyboard when a custom keyboard is used exclusively.


AntonioND:
Thanks!

PR #93, opened 2024-04-27
fat: nitrofs: Implement get/set FAT file attribute functions

AntonioND:
This is done for compatibility with libfat. It may be useful in some situations.

Note that in NitroFS it is only possible to get file attributes, as it is a read-only filesystem.

### DIFF
--- a/source/arm9/libc/filesystem.c
+++ b/source/arm9/libc/filesystem.c
@@ -652,3 +652,41 @@ int symlink(const char *target, const char *path)
     return -1;
 }
 
+int FAT_getAttr(const char *file)
+{
+    if (nitrofs_use_for_path(file))
+        return nitrofs_fat_get_attr(file);
+
+    FILINFO fno = { 0 };
+    FRESULT result = f_stat(file, &fno);
+
+    if (result != FR_OK)
+    {
+        errno = fatfs_error_to_posix(result);
+        return -1;
+    }
+
+    return fno.fattrib;
+}
+
+int FAT_setAttr(const char *file, uint8_t attr)
+{
+    if (nitrofs_use_for_path(file))
+    {
+        errno = EROFS; // Read-only filesystem
+        return -1;
+    }
+
+    // Modify all attributes (except for directory and volume)
+    BYTE mask = AM_RDO | AM_ARC | AM_SYS | AM_HID;
+
+    FRESULT result = f_chmod(file, attr, mask);
+
+    if (result != FR_OK)
+    {
+        errno = fatfs_error_to_posix(result);
+        return -1;
+    }
+
+    return 0;
+}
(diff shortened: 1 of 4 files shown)


source/arm9/libc/nitrofs.c
```diff
@@ -521,6 +521,28 @@ static int nitrofs_stat_file_internal(nitrofs_file_t *f, struct stat *st)
...
+    {
+        errno = ENOENT;
+        return -1;
+    }
+    else if (res >= 0xF000)
+    {
+        return ATTR_DIRECTORY | ATTR_READONLY;
+    }
+
+    return ATTR_READONLY;
```
AntonioND, line 543: Not sure about "ATTR_ARCHIVE".

asiekierka: Windows:

> When set, it indicates that the hosting file has changed since the last backup operation. Windows' file system sets this attribute on any file that has changed.

I think we can safely skip it here.

source/arm9/libc/nitrofs.c
```diff
@@ -521,6 +521,28 @@ static int nitrofs_stat_file_internal(nitrofs_file_t *f, struct stat *st)
...
+        return -1;
+    }
+
+    int32_t res = nitrofs_path_resolve(name);
+    if (res < 0)
+    {
+        errno = ENOENT;
+        return -1;
+    }
+    else if (res >= 0xF000)
```
asiekierka, line 538: We don't need an `else` here; the above `if`-block returns.

asiekierka:
Other than the comments, LGTM.

source/arm9/libc/filesystem.c
```diff
@@ -652,3 +652,38 @@ int symlink(const char *target, const char *path)
...
+        return -1;
+    }
+
+    return fno.fattrib;
+}
+
+int FAT_setAttr(const char *file, uint8_t attr)
+{
+    if (nitrofs_use_for_path(file))
+        return -1;
```
asiekierka, line 675: Does `libfat` actually use `errno` for the error codes?

Either way, if we're using `errno`, it might be a good idea to set it to something here as well.

source/arm9/libc/nitrofs.c
AntonioND: Yeah, that was my thought too, but I wanted an unbiased opinion. Thanks.

source/arm9/libc/filesystem.c
asiekierka: To clarify: I don't think we should aim to be quirk-for-quirk compatible with `libfat`; this seems like a more consistent/harder to get wrong approach. I was just curious.

AntonioND: It doesn't, I think? But just getting "-1" is super annoying if you're trying to debug something, and it's essentially free to support errno in our code, so we might as well do it.

asiekierka: Yes, in that case it's best to set a sensible `errno` for all error paths.

AntonioND: Added.

source/arm9/libc/nitrofs.c
AntonioND: Fixed

PR #94, opened 2024-04-29
Add hardware accelerated floating point squareroot

Kuratius:
Result should be bit-for bit identical to sqrtf from the standard math library, but runs 4x faster.
Currently doesnt handle subnormals (Numbers smaller than 2^-126, e.g. 0.9*2^-127, are called subnormal because they don't follow the standard floating point format, which messes up the math, will add this to the warning.)

Typically subnormal numbers are disabled via -ffast-math anyway.

### DIFF
--- a/include/nds/arm9/math.h
+++ b/include/nds/arm9/math.h
@@ -319,6 +319,15 @@ static inline uint32_t sqrt64(uint64_t a)
     return sqrt64_result();
 }
 
+/// 32-bit floating point sqrt
+///
+/// @warning Subnormal floats are treated as if they were zero.
+/// @warning Not safe to call inside an interrupt handler.
+/// 
+/// @param valid 32 bit non-negative floating point value.
+/// @return 32 bit floating point value
+ARM_CODE float hw_sqrtf(float x);
+
 /// 20.12 fixed point cross product.
 ///
 /// Cross product:
--- a/source/arm9/math.c
+++ b/source/arm9/math.c
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: Zlib
+//
+// Copyright (C) 2024 Dominik Kurz
+#include <nds/arm9/math.h>
+#define QUIET_NAN ((255 << 23) | ( ( 1 << 22 ) | 1 ))
+#define INF (0xFF << 23)
+
+ARM_CODE float hw_sqrtf(float x)
+{
+    union{float f; u32 i;}xu;
+    xu.f = x;
+    u32 mantissa32 = xu.i & ((1 << 23) - 1); // extracts mantissa via bitmask
+    mantissa32 += (1 << 23); // adds implicit bit to mantissa.
+    // If you were planning to handle subnormal numbers 
+    // you would check if the number is subnormal before 
+    // adding the implicit bit.
+    // You would also need to do the shift of the mantissa and exponent
+    // differently, probably should use __builtin_clz for this
+    // as the count leading zeroes (clz) instruction is 1 cycle on arm
+    // but this is not my problem
+    // performance-critical code has subnormals turned off
+    // due to -ffast-math 
+    u64 mantissa = mantissa32; // cast so that the shift doesnt go wrong 
+    mantissa <<= (((xu.i & (1 << 23)) == 0) + 25);
+    // applies additional shift depending on whether exponent is even or odd
+    // 25 is 23+2, the +2 is necessary for rounding
+    REG_SQRTCNT = SQRT_64;  
+    REG_SQRT_PARAM = mantissa;
+    // starts hardware squareroot 
+    // It is critical that this happens as early as possible so that
+    // we have time to do other stuff in the meantime
+    u32 raw_exponent = (xu.i & (255 << 23));
+    u32 sign = (xu.i & (1 << 31));
+    // check if exponent is 0
+    if (raw_exponent != 0)  
+    {
+        if (sign  || (xu.i > (255 << 23))) //check if negative or NaN
+        {
+            // Expected behavior:
+            // sqrt(-f)=+qNaN, sqrt(-NaN)=+qNaN,sqrt(-Inf)=+qNaN 
+            // The IEEE 754 standard doesnt specify how signalling NaN
+            // is encoded. Hard float arm stuff does it with a 1 in the least
+            // significant bit and zeros in the rest of the mantissa
+            // Quiet NaNs  on ARM are a 1 in 
+            // the most and least significant bit of the mantissa.
+            // Example:
+            // Quiet NaN mantissa (100000.....00001)
+            // Signalling NaN mantissa (00000.....00001)
+            xu.i= QUIET_NAN;
+            while(REG_SQRTCNT & SQRT_BUSY); // wait for sqrt
+            return xu.f;
+        } else if (raw_exponent == (255  << 23)) // check for +Inf
+        {
+	    xu.i=INF;
+            while(REG_SQRTCNT & SQRT_BUSY); // wait for sqrt 
+            return xu.f; // return Inf	 
+        }
+        s32 exponent = raw_exponent - (127 << 23);  
+        // subtract bias from exponent
+        exponent >>= 1;
+        // This is meant to be a floor division
+        // meaning -1/2= -0.5 should map to -1
+        // if your compiler does logical shifts instead of arithmetic 
+        // use a division by 2.
+        // arithmetic shift and division have different rounding behavior 
+        // but in this case it doesnt matter
+        exponent = ((exponent + (127 << 23)) & (0xff << 23) ); 
+        // add bias again and bitmask the exponent
+        while(REG_SQRTCNT & SQRT_BUSY); 
+        // waits for square root operation to complete
+        u32 new_mantissa=REG_SQRT_RESULT;
+        new_mantissa += 1; // necessary for rounding
+        new_mantissa >>= 1;// shift down after rounding 
+        // construct float from exponent and mantissa
+        xu.i =  exponent| (new_mantissa & ((1 << 23) - 1)); 
+        return xu.f;
+    } else
+    {
+        // numbers in this branch are either 0 or subnormal
+        // if someone wants to add subnormal number handling 
+        // this would be the branch to do it
+        // right now it assumes subnormal numbers are 0.
+        // the --fast-math option causes other functions to generate 0
+        // if they would otherwise generate a subnormal number
+        // so maybe make sure subnormal number handling is done with a define
+        // that checks if this option is present
+        xu.i=sign;
+        while(REG_SQRTCNT & SQRT_BUSY);// wait for sqrt to be done
+        return xu.f; //returns +0 or -0 as appropriate
+    }
+
+}
+


include/nds/arm9/math.h
```diff
@@ -319,6 +319,14 @@ static inline uint32_t sqrt64(uint64_t a)
     return sqrt64_result();
 }
 
+
+/// 32 bit floating point square root
+//  @param valid 32 bit positive floating point value. NaN and Inf are not handled.
```
AntonioND, line 324: The comments need to start with three slashes in all lines or Doxygen won't see them. Also, leave an empty line with only three slashes right below the description.

Also, the note about NaN and Inf is important, you should use `@warning` so that it appears very clearly in the doxygen documentation.

source/arm9/hw_math.c
```diff
@@ -0,0 +1,35 @@
+
+
+#include <math.h>
+
+
+
+f32 hw_sqrtf(f32 x){
```
AntonioND, line 7: `f32` is a libnds type, right? I think using `float` makes more sense, in order to make it clear that it's a simple float. Using types like `int32_t` and such is fine, probably.

Also, the bracket needs to go in the next line.

source/arm9/hw_math.c
```diff
@@ -0,0 +1,35 @@
...
+	exponent=((exponent+(127<<23))& (0xff<<23) );
+    //fetch async result here
+	while(REG_SQRTCNT & SQRT_BUSY);
+	u32 new_mantissa=REG_SQRT_RESULT;
+	new_mantissa+=1;
+	new_mantissa>>=1;
+	xu.i=  exponent| (new_mantissa & ((1<<23)-1));
+	return xu.f;
+    } else{
+	while(REG_SQRTCNT & SQRT_BUSY);
```
AntonioND, line 30: Do you really need to wait for the hardware to not be busy? It's probably okay to leave it running. If the user wants to start a new operation they will overwrite the values in the registers, they don't have to wait until the hardware is done (I may be wrong here, though).

source/arm9/hw_math.c
```diff
@@ -0,0 +1,35 @@
...
+	return 0.0;
+    };
```
AntonioND, line 32: That semicolon isn't required.

source/arm9/hw_math.c
```diff
@@ -0,0 +1,35 @@
...
+    xu.f=x;
+    u64 mantissa=xu.i & ((1<<23)-1);
+    //check if exponent is odd
+    //before subtracting 127 exponent was even if it is odd now
+    //therefore check if last digit is 0 
+    mantissa=(mantissa+(1<<23))<<   ( (( xu.i & (1<<23))==0  )      +25 );
+    REG_SQRTCNT = SQRT_64;
+    REG_SQRT_PARAM = mantissa;
+    u32 raw_exponent= (xu.i & (0xff<<23));
+    if (raw_exponent > 0 ) {
```
AntonioND, line 18: In general, could you try to imitate the code style of the rest of the library? For example, don't leave that space after the 0, don't leave that curly bracket in that line (move it to the next one). Leave one space after `//` in your comments, always leave one space on both sides of a `=` or other operators.

source/arm9/hw_math.c
```diff
@@ -0,0 +1,35 @@
+
```
AntonioND, line 1: You need to add a license header. Copy one from a different file. Also, since you're adding the declaration to `math.h`, rename this to `math.c` for consistency.

source/arm9/hw_math.c
```diff
@@ -0,0 +1,35 @@
...
+
+
+f32 hw_sqrtf(f32 x){
+    union{f32 f; u32 i;}xu;
+    xu.f=x;
+    u64 mantissa=xu.i & ((1<<23)-1);
+    //check if exponent is odd
+    //before subtracting 127 exponent was even if it is odd now
+    //therefore check if last digit is 0
```
AntonioND, line 13: The comment isn't clear.

AntonioND (changes requested):

source/arm9/hw_math.c
Kuratius: I was under the impression that REG_SQRCNT can be both written and read to, and so if the sqrt unit changes its value after the first sqrt is done it might lead to fetching the result of the previous sqrt operation.

source/arm9/hw_math.c
Kuratius: Added one, let me know if I should use a different one instead. I'm unsure as to the difference between zlibg and bsd licenses.

Kuratius:
I should probably also change the f32 in line 9 to float

And the stuff in line 11,12 only needs a 64 variable for the shift, the addition doesn't need to be 64 bit so it probably should be in a separate line.
Probably only affects speed in thumb mode though, as I think 64 bit adds are similar or the same speed as 32 bit adds.

include/nds/arm9/math.h
```diff
@@ -319,6 +319,16 @@ static inline uint32_t sqrt64(uint64_t a)
     return sqrt64_result();
 }
 
+
+/// 32 bit floating point square root
+///  @param valid 32 bit positive floating point value.
+///  @warning Subnormals are treated as if they were zero.  NaN and Inf are not handled. 
+///  @return 32 bit floating point value
+///
```
AntonioND, line 327:
```
/// 32 bit floating point square root.
///
/// @warning Subnormals are treated as if they were zero. NaN and Inf are not handled. 
///
/// @param valid 32 bit positive floating point value.
/// @return 32 bit floating point value.
float hw_sqrtf(float x);
```

Don't add empty lines before the comment, and don't add many lines after the function declaration, just one.

Also, I'm not supre sure about the name of the function, but I don't have good ideas. If someone wants to propose something...

source/arm9/math.c
```diff
@@ -0,0 +1,38 @@
+
```
AntonioND, line 1: Remove the empty line on top, and follow the pattern of the other files (leave the empty comment in the middle with just "//". Also, leave an empty line after the comments and before the include.

source/arm9/math.c
```diff
@@ -0,0 +1,38 @@
...
+    union{f32 f; u32 i;}xu;
+    xu.f = x;
+    u64 mantissa = xu.i & ((1 << 23) - 1); // extracts mantissa via bitmask
+    mantissa = (mantissa + (1 << 23)) << (((xu.i & (1<<23)) == 0) + 25); // applies additional shift depending on whether exponent is even or odd
+    REG_SQRTCNT = SQRT_64;  
+    REG_SQRT_PARAM = mantissa;
+    u32 raw_exponent = (xu.i & (0xff << 23));
+    // check if exponent is 0
+    if (raw_exponent > 0) 
+    {
```
AntonioND, line 18: You're using tabs and spaces in the same function, replace everything by spaces. In here I don't see indentation correctly, for example. And people may have set up tabs to be 4 spaces, not 8 (which I think is your setup).

source/arm9/math.c
```diff
@@ -0,0 +1,38 @@
...
+// SPDX-License-Identifier: Zlib
+// Copyright (C) 2024 Dominik Kurz
+#include <math.h>
+
+
+float hw_sqrtf(float x)
+{
+    union{f32 f; u32 i;}xu;
+    xu.f = x;
+    u64 mantissa = xu.i & ((1 << 23) - 1); // extracts mantissa via bitmask
```
AntonioND, line 11: Some of your lines are too long. Try to keep them at most at 80 characters if possible, and definitely at less than 100 characters (for people that have two files open side by side).

source/arm9/hw_math.c
AntonioND: Zlib, so that it's the same as the rest of the library.

source/arm9/hw_math.c
AntonioND: Hmmm well, I don't know, let's leave it like this for now, it won't hurt.

Kuratius:
Also, do you want me to add my tests for this function somewhere?

Kuratius:
Also I just realized this might be the wrong header;
it should probably be <nds/arm9/math.h> as I think <math.h> will probably take the C standard library header.

source/arm9/math.c
```diff
@@ -0,0 +1,49 @@
...
+    REG_SQRT_PARAM = mantissa;
+    // starts hardware squareroot 
+    u32 raw_exponent = (xu.i & (0xff << 23));
+    // check if exponent is 0
+    if (raw_exponent > 0) 
+    {
+        s32 exponent = raw_exponent-(127<<23);  // subtract bias from exponent
+        exponent = exponent >> 1; 
+        // right shift on signed number depends on compiler
+        // but it works in gcc
```
asiekierka, line 28: I am skeptical of this, especially as I test with Clang from time to time. If it's a GCC-ism, it should be put behind a GCC-specific `#ifdef`.

Kuratius: I found a comment on stack exchange stating that arithmetic right shift and integer division have different rounding behaviours, so maybe division wouldn't work.  Would probably have to test. Wikipedia says almost every compiler does arithmetic shifts though, so it'll probably work in clang.

Kuratius: image
seems /2 gets converted to an ASR instruction on nds

AntonioND: To be fair, we already rely on the fact that ">>" on signed integers is an arithmetic shift in other places, like the fixed point math functions. I'm pretty sure this is the kind of things that Clang would never implement in a different way than GCC.

Kuratius: I'll change it back to a shift then.  There are cases where gcc will emit division instructions with division by 2 for some optimization levels and for some numbers. I'll just explain it in a comment.

Kuratius:
should now preserve NaN and Inf

AntonioND:
I've created this test. It looks like this function is indeed pretty fast, maybe twice as fast as the software version (in my DSi I get 152 ticks per op for software, 94 ticks for hardware accelerated): https://github.com/blocksds/sdk/tree/195e903952bf1c4fbb623c2ad91bf48c2d5b2312/tests/system/hw_math

I won't have time to merge the PR until Monday, though.

Kuratius:
> I've created this test. It looks like this function is indeed pretty fast, maybe twice as fast as the software version (in my DSi I get 152 ticks per test iteration for software-based, 94 ticks for hardware-accelerated): https://github.com/blocksds/sdk/tree/195e903952bf1c4fbb623c2ad91bf48c2d5b2312/tests/system/hw_math
> 
> I won't have time to merge the PR until Monday, though. Feel free to improve the function until then, or to leave comments.
> 
> Also, maybe it's worth documenting that negative numbers aren't handled (yes, the comment mentions NaNs, but it's not very clear that you can get one from a negative float).

You're benchmarking float multiplies and square roots together here.

AntonioND:
Ah, right! I'll fix that on Monday.

Kuratius:
I looked up how signalling NaNs are encoded and added a branch for it.
But I think it is not ideal to do that as it means that people now need to take the absolute before passing it to the function, so for now it's commented out.

Kuratius:
https://www.felixcloutier.com/x86/fsqrt

This seems to indicate that fsqrt instructions typically just call an exception instead of returning a NaN on negative non-zero input.

https://github.com/rofl0r/musl/blob/master/src/math/sqrtf.c

This seems to return a NaN on negative inputs.
So it seems like that isnt so clearly defined.

asiekierka:
> This seems to indicate that fsqrt instructions typically just call an exception instead of returning a NaN on negative non-zero input.

What a processor does is irrelevant to the C standard. The only thing relevant to the C standard is the specification, of which the final draft (essentially equal to the paid document) is our best source. The latest one, for C23, is [here](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3088.pdf) and states:

> The sqrt functions compute the nonnegative square root of x. A domain error occurs if the argument is less than zero.

A domain error's behaviour is defined (or rather, not defined) as follows:

> Whether a signaling NaN input causes a domain error is implementation-defined. On a domain error, the function returns an implementation-defined value [...]

Kuratius:
https://pubs.opengroup.org/onlinepubs/007904975/functions/sqrt.html
```For finite values of x < -0, a domain error shall occur, and [[MX](javascript:open_code('MX'))] [Option Start]  either a NaN (if supported), or [Option End]  an implementation-defined value shall be returned.```

File: source/arm9/math.c [hunk 13]
```diff
@@ -0,0 +1,101 @@
...
+        // Honestly there is probably a better way to do this.
+        // as the IEEE 754 standard doesnt specify how signalling NaN
+        // is encoded. Hard float arm stuff does it with a 1 in the least
+        // significant bit and zeros in the rest of the mantissa
+        // Quiet NaNs  on ARM are a 1 in both 
+        // the most and least significant bit of the mantissa.
+        // Example:
+        // Quiet NaN mantissa (100000.....00001)
+        // Signalling NaN mantissa (00000.....00001)
+            xu.i= (255 << 23) |( ( 1 << 22 ) |1 );
```
AntonioND, line 57: Hardware ARM implementations may leave a 1 in the LSB, but the software implementation of libm leaves a 0 there... I'm not sure what's the right behaviour here.

AntonioND:
I've improved my test to fix the profiling issue and to test for NaNs: https://github.com/blocksds/sdk/blob/5de2b3bbad6e79a94c9a8b84870ca62eafea53f6/tests/system/hw_math/source/main.c

Kuratius:
Your signalling nan in the test is Inf according to Wikipedia.

source/arm9/math.c
Kuratius: Probably not important as long as it is a type of NaN. It's mostly to help people find bugs I think, returning NaN isn't that useful beyond that.

AntonioND:
True. What confused me was this other Wikipedia article:

https://en.m.wikipedia.org/wiki/NaN

> For example, an IEEE 754 single precision (32-bit) NaN would be encoded as
>
>     s111 1111 1xxx xxxx xxxx xxxx xxxx xxxx
>
> where s is the sign (most often ignored in applications) and the x sequence represents a non-zero number (the value zero encodes infinities). In practice, the most significant bit from x is used to determine the type of NaN: "[quiet NaN](https://en.m.wikipedia.org/wiki/NaN#Quiet_NaN)" or "[signaling NaN](https://en.m.wikipedia.org/wiki/NaN#Signaling_NaN)" (see details in [Encoding](https://en.m.wikipedia.org/wiki/NaN#Encoding)). The remaining bits encode a payload (most often ignored in applications).

It mentions it has to be non-zero, but I skipped over it.

AntonioND:
So one last thing. Can you squash all your commits and rebase them on top of the current `master` branch (with `git rebase -i origin/master`, for example)?

Also, rename the commit to `math: Add hardware accelerated floating point sqrt`. I could do this, but I'd rather leave the author of the commit untouched.

source/arm9/math.c
AntonioND: Sure.

In any case, the `sqrtf()` of `math.h` returns a quiet NaN (`got` is the `hw_sqrt` version, `expected` is the `math.h` version):

image

Maybe it's better for them to be the same?

source/arm9/math.c
```diff
@@ -0,0 +1,101 @@
...
+    // starts hardware squareroot 
+    // It is critical that this happens as early as possible so that
+    // we have time to do other stuff in the meantime
+    u32 raw_exponent = (xu.i & (0xff << 23));
+    u32 sign = (xu.i & (1 << 31));
+    // check if exponent is 0
+    if (raw_exponent != 0)  
+    {
+        if (sign) //check if negative
+        {
```
AntonioND, line 37: While you're at it, can you indent the comment below 4 more spaces to the right? Don't create a new commit for that, just squash all commits and leave one only commit with the title I've mentioned.

source/arm9/math.c
```diff
@@ -0,0 +1,101 @@
+// SPDX-License-Identifier: Zlib
+//
+// Copyright (C) 2024 Dominik Kurz
+#include <nds/arm9/math.h>
+
+
+float hw_sqrtf(float x)
```
AntonioND, line 7: I've experimented a bit with this. Can you add `ARM_CODE` in front of this? It seems to save a few CPU cycles. Also add `ARM_CODE` in the header:
```
ARM_CODE float hw_sqrtf(float x)
```

source/arm9/math.c
Kuratius: You're testing a sNaN as input, so it preserves it.
The branch you quoted is for negative numbers, in which case they are the both quiet NaNs.
image

Kuratius: Actually nevermind, I can probably just delete the bit that makes a NaN signalling.

AntonioND: > You're testing a sNaN as input, so it preserves it. The branch you quoted is for negative numbers, in which case they should both be quiet NaNs.

According to this: https://en.wikipedia.org/wiki/NaN

> Signaling NaNs, or sNaNs, are special forms of a NaN that, when consumed by most operations, should raise the invalid operation exception and then, if appropriate, be "quieted" into a qNaN that may then propagate.

So it looks like `math.h` is doing the right thing by turning the sNaN into a qNaN.

Kuratius:
This is probably messed up now, probably dont merge this until I figure out how to fix the git issues.

Kuratius:
I think it's ok now?

Kuratius:
If this compiles now I'm not gonna touch it anymore.

Kuratius:
Aaand I messed up a previous commit

AntonioND:
Ok, it still fails two corner cases (the NaN has bit 0 set instead of unset, which is very minor) so I'm going to merge it. Thanks, and sorry for all the hassle.

AntonioND:
I've also merged my tests: https://github.com/blocksds/sdk/commit/ace1f47eb4a4cbf37385a99285478d238450920d

Kuratius:
Nice!
Btw, I noticed that
if (sign || (xu.i > (255 << 23))) is logically equivalent to I assume a common use case for a faster sqrt is probably  if (xu.i > (255 << 23))  due to xu.i being an unsigned int and sign being the first bit of xu.i, but it's probably not worth changing as it wouldn't affect performance because it's done while async anyway.

Just writing it so that it's documented.

PR #95, opened 2024-05-01
arm9: Fix some memory allocation issues found by GCC 14.

asiekierka:


source/arm9/libc/chdir.c
```diff
@@ -101,7 +101,7 @@ char *getcwd(char *buf, size_t size)
             optimize_mem = 1;
         }
 
-        buf = calloc(size, 1);
+        buf = malloc(size);
```
asiekierka, line 104: This can just be a `malloc`, as I think we can trust `getcwd` to return a valid string if no error has been raised?

AntonioND: Sounds good.

AntonioND:
Thanks!

PR #96, opened 2024-05-07
include: rewrite sys/dirent.h, fix minor issues

asiekierka:
I created a new `sys/dirent.h` by deleting the old one and writing a new one based on [the POSIX specification](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/), as well as [the GNU libc manual](https://www.gnu.org/software/libc/manual/html_node/Directory-Entries.html) and other manuals for BSD/GNU-specific extensions.

This PR means we can finally remove the last disputed bit of BSD/MIT-style licensing. This PR also fixes some minor issues, including but possibly not limited to:

- `d_ino` is now properly documented, and its (and `d_off`)'s sizes are controlled by us (in `machine/_types.h`) to match BlocksDS's requirements,
- `MAXNAMLEN` is now set to the correct maximum value that FatFS can output.

### DIFF
--- a/include/machine/_types.h
+++ b/include/machine/_types.h
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: Zlib
+//
+// Copyright (c) 2024 Adrian "asie" Siekierka
+
+#ifndef MACHINE__TYPES_H__
+#define MACHINE__TYPES_H__
+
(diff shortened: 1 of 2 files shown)


AntonioND:
Cool, thanks!

PR #97, opened 2024-05-29
Start Division/Sqrt earlier on repeated calls

Kuratius:
Possibly comes at a slight performance cost if someone is constantly switching modes, but I don't think that is a common usage pattern. Still haven't figured out how to use stmia for writing the arguments of a division, but probably isn't worth the trouble. Probably better to only use stmia for a 64 bit by 64 bit divide (this mode is supported but doesnt have a helper function yet).

Might need some testing, don't merge this immediately.


asiekierka:
> Still haven't figured out how to use stmia for writing the arguments of a division

The way that typically works for me is: use a struct, then assign a struct to another struct. [Example.](https://github.com/asiekierka/uxnds/blob/main/arm9/source/nds_ppu.c#L402-L411) Alternatively, try rewriting these routines in inline assembly for full control.

Kuratius:
```inline void start_div64_64(s64 a, s64 b){
    register uint loa asm("r0") =a;;
    register uint hia asm("r1")=a>>32;
    register uint lob asm("r2")=b;
    register uint hib asm("r3")=b>>32;
asm volatile(
    "mov r12, #0x04000000 ; \n\t"
    "add r12, r12, #0x290; \n\t"
    "stmia r12, {r0-r3}; \n\t"
    "ldr r2, [r12, #-0x10]; \n\t"
    "mov r3, #2; \n\t"
    "and r2, r2, #3; \n\t"
    "cmp r2, #2; \n\t"
    "strne r3, [r12, #-0x10]; \n"
    : [loa]"+r"(loa), [hia]"+r"(hia), [lob]"+r"(lob), [hib]"+r"(hib)::    "r12", "cc"
    );
}

inline s64 finish_div64_64(){
    register uint lo asm("r0");
    register uint hi asm("r1");
asm volatile(
    "mov %[hi], #0x04000000;\n\t"
    "add %[hi], %[hi], #0x290;\n\t"
    "ldr %[lo], [%[hi], #-0x10];\n"
"1:\n\t"
    "tst %[lo], #0x8000;\n\t"
    "ldrne %[lo], [%[hi], #-0x10];\n\t"
    "bne 1b ;\n\t"
    "ldrd %[lo], %[hi],[%[hi], #0x10];\n"
    : [lo]"=r"(lo), [hi]"=r"(hi) :: "cc"
    );
    return ((u64) hi<<32) | lo  ; 
}``` 

This uses gcc inline assembly syntax, so I'm not sure if it'll work with clang, This also forces the compiler to store the arguments in r0-r3, this results in a few extra mov instructions for the starting function being added. You can avoid forcing use of specific registers, but then using stm is not possible anymore unless there is a way to make sure they are in ascending order.

The "and" and "cmp" in the starting function can be replaced with a tst to make the starting function even shorter,, but this might not be safe if someone writes garbage to the control register, i.e. it would result in wrong behavior if the control is ever (11).

File: include/nds/arm9/math.h [hunk 1]
```diff
@@ -106,8 +109,11 @@ static inline int32_t mulf32(int32_t a, int32_t b)
 /// @param a 20.12 positive value.
 static inline void sqrtf32_asynch(int32_t a)
 {
-    REG_SQRTCNT = SQRT_64;
     REG_SQRT_PARAM = ((uint64_t)(uint32_t)a) << 12;
+    if ((REG_SQRTCNT & 0xF) != SQRT_64)
```
AntonioND, line 113: According to GBATEK, only bit 0 determines the mode for square roots. Why are you using `0xF` as a mask? shouldn't it just be `BIT(0)`? https://problemkaputt.de/gbatek.htm#dsmaths

For the division it's the lowest two bits. Maybe we need to add new defines to `math.h`:

```diff
 #define DIV_64_64           2
 #define DIV_64_32           1
 #define DIV_32_32           0
+#define DIV_MODE_MASK       3
+
 #define DIV_BUSY            (1 << 15)

 #define SQRT_64             1
 #define SQRT_32             0
+#define SQRT_MODE_MASK      1
+
 #define SQRT_BUSY           (1 << 15)
```

And then use `DIV_MODE_MASK` and `SQRT_MODE_MASK`.

AntonioND (changes requested):

include/nds/arm9/math.h
Kuratius: I was using 0xF because it was convenient and it works, it's probably not optimal. Your suggestion of checking one bit is probably 1 instruction less (better) because it should generate a tst instruction instead of AND+CMP so in the unlucky case  where we are in the wrong mode the cnt write will happen faster.

We could technically do the same for div, but as I mentioned for div that relies on nobody trying to write garbage to the register.

AntonioND: > I was using 0xF because it was convenient and it works because the extra bits are unused, it's probably not optimal. Your suggestion of checking one bit is probably 1 instruction less (better) because it should generate a tst instruction instead of AND+CMP so in the unlucky case where we are in the wrong mode the cnt write will happen faster.
> 
> We could technically do the same for div, but as I mentioned for div that relies on nobody trying to write garbage to the register as (11) would result in wrong behaviour.

Sure, but my proposal is using 3 as mask. It would mean doing something like this:
```
if ((REG_DIVCNT & DIV_MODE_MASK) != DIV_64_32)
...
if ((REG_SQRTCNT & SQRT_MODE_MASK) != SQRT_64)
```

Kuratius: I've added these changes now.

source/arm9/math.c
```diff
@@ -93,7 +94,15 @@ ARM_CODE float hw_sqrtf(float x)
         // so maybe make sure subnormal number handling is done with a define
         // that checks if this option is present
         xu.i = sign;
-        while (REG_SQRTCNT & SQRT_BUSY);// wait for sqrt to be done
         return xu.f; // returns +0 or -0 as appropriate
     }
 }
+
```
AntonioND, line 100: Can you remove this extra whitespace?

Kuratius: Done.

include/nds/arm9/math.h
```diff
@@ -64,9 +65,12 @@ extern "C" {
 /// @param den 20.12 denominator.
 static inline void divf32_asynch(int32_t num, int32_t den)
 {
-    REG_DIVCNT = DIV_64_32;
     REG_DIV_NUMER = ((int64_t)num) << 12;
     REG_DIV_DENOM_L = den;
+    if ((REG_DIVCNT & DIV_MODE_MASK) != DIV_64_32)
+    {
+	REG_DIVCNT = DIV_64_32;
```
AntonioND, line 72: Don't use tabs.

include/nds/arm9/math.h
```diff
@@ -44,11 +44,12 @@ extern "C" {
 #define DIV_64_32           1
 #define DIV_32_32           0
 #define DIV_BUSY            (1 << 15)
+#define DIV_MODE_MASK       3
 
 #define SQRT_64             1
 #define SQRT_32             0
 #define SQRT_BUSY           (1 << 15)
-
+#define SQRT_MODE_MASK      1
```
AntonioND, line 52: Leave an empty line between the defines and the comment.

include/nds/arm9/math.h
```diff
@@ -214,7 +230,6 @@ static inline void div64_asynch(int64_t num, int32_t den)
 static inline int32_t div64_result(void)
 {
     while (REG_DIVCNT & DIV_BUSY);
-
```
AntonioND, line 217: Don't change whitespace randomly in other parts of the code, it makes it harder to analyze the history of the file.

AntonioND:
Squashed and merged here: https://github.com/blocksds/libnds/commit/72f444cc4a6b2025ee52c44dd7e39e5f877d1ff5

PR #101, opened 2024-06-09
fatfs: update wf-fatfs fork

asiekierka:
Added some slight performance optimizations to directory searches; saves ~100-200 bytes of code, too... however, I felt it wasn't necessarily a good idea to merge this immediately before a new release.


AntonioND:
Thanks!

PR #102, opened 2024-06-13
arm7: Rewrite touch code, refactor SPI, RTC code

asiekierka:
- Rewrote touch code.
  - A shared, from-scratch measurement filtering routine is now used for both TSC (NTR) and CDC (TWL) touch inputs.
  - On TSC, tscMeasure() uses the 16-clock-per-conversion method to speed up measurement readouts.
  - On both TSC and CDC, critical section code has been reduced to exclude post-measurement filtering, improving ARM7 responsiveness.
- Added helper SPI_TARGET defines and spiRead, spiWrite, spiExchange functions to clean up SPI communication code.
  - Renamed SerialWaitBusy to spiWaitBusy to match.
- Cleaned up microphone input code to use tscRead().
- Various header cleanups:
  - Firmware flash commands have been documented in firmware.h, which was split out from serial.h.
  - NTR touch screen controller commands have been documented in tsc.h, which was split out from touch.h.
  - RTC_CR has been renamed to REG_RTCCNT.
  - RTC bus control CS/SCK/IO defines have been moved from internal to public code.
  - serial.h has been split out into firmware.h.
  - touch.h has been split out into tsc.h.

The new algorithm still needs work.

### DIFF
--- a/include/nds.h
+++ b/include/nds.h
@@ -96,7 +96,11 @@
 /// - @ref nds/arm7/clock.h "RTC utilities"
 /// - @ref nds/arm7/input.h "Keypad and touch pad ARM7 helpers"
 /// - @ref nds/arm7/audio.h "Audio and microphone helpers"
-/// - @ref nds/arm7/tmio.h "TMIO ARM7 module"
+/// - @ref nds/arm7/touch.h "Touch screen helpers"
+/// - @ref nds/arm7/serial.h "SPI bus controller"
+/// - @ref nds/arm7/firmware.h "Firmware flash access helpers"
+/// - @ref nds/arm7/tsc.h "DS Touchscreen/Sound controller helpers"
+/// - @ref nds/arm7/tmio.h "DSi TMIO ARM7 module"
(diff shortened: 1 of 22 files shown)


asiekierka:
Testing steps:

1. Download https://asie.pl/files/touch_input_20240614_2226.nds
2. Press RIGHT once to show the "Adjusted input" mode on the touchscreen, so you can see your touch input being tracked as you draw.
3. Take note of the Z1/Z2 values as you glide along/adjust stylus strength. In this build, they are repurposed to refer to the amount of noise on the X and Y axis. (For example, on my worn DS Lite, the difference is <10 samples, while in a specific area on the left border it goes up to 15-20 samples. For context, ~15 samples = ~1 pixel of noise.)
4. Let me know!

AntonioND:
As discussed in private, it looks like preliminary testing shows that this works better than before, and you can send improvements to the driver as time passes. I've merged it so that we don't have this PR hanging for too long and it gets tested by a wider audience.

PR #104, opened 2024-06-14
interrupts: new dispatcher

asiekierka:
- Constant-time IRQ lookups (always faster on ARM9, faster for second IRQ onwards on ARM7)
- Reduced IRQ table size (saves 128 bytes on ARM9, 188 bytes on ARM7)

Also fixes a bug in `irqClearAUX`.

### DIFF
--- a/source/common/interruptDispatcher.s
+++ b/source/common/interruptDispatcher.s
@@ -30,8 +30,13 @@ BEGIN_ASM_FUNC IntrMain
     stmfd   sp!, {r0-r1, r12, lr}   // {spsr, IME, REG_BASE, lr_irq}
 
     add     r12, r12, #0x210
-    ldmia   r12, {r1, r2}
+    ldmia   r12!, {r1, r2}
     ands    r1, r1, r2
+#ifdef ARM9
+    beq     no_handler
+#else
+    beq     setflagsaux
+#endif
     ldr     r2, =irqTable
 
     // Notify the BIOS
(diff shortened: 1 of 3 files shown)


source/common/interruptDispatcher.s
```diff
@@ -67,46 +74,65 @@ BEGIN_ASM_FUNC IntrMain
...
+    // r2 = target irq table address
+    // r0, r3 can be used freely
 findIRQ:
-
-    ldr     r0, [r2, #4]
-    cmp     r0, #0
+#ifdef ARM9
+    clz     r0, r1
+    eor     r0, r0, #31
+#else
```
AntonioND, line 90: Can you add a comment here explaining a bit what the code below is doing? Is it just the equivalent of `clz`?

source/common/interruptDispatcher.s
```diff
@@ -67,46 +74,65 @@ BEGIN_ASM_FUNC IntrMain
...
-    b       no_handler
 
 got_handler:
 
-    str     r0, [r12, #4]           // IF Clear
-
     mrs     r2, cpsr
-    mov     r3, r2
-    bic     r3, r3, #0xdf           // Enable IRQ & FIQ. Set CPU mode to System
+    bic     r3, r2, #0xdf           // Enable IRQ & FIQ. Set CPU mode to System
```
AntonioND, line 125: Can you use the `CPSR` defines in `include/nds/cpu_asm.h` instead of having this number here? I guess you could also replace the magic number below, that's up to you.

source/common/interruptDispatcher.s
asiekierka: The `clz`/`eor` and ARM7 code are equivalent; it's just that the `clz` instruction is ARM9-exclusive.

AntonioND:
Okay, I get it, this turns the array of structs of arbitrary handlers (so that you can assign more than one handler to each instruction) to a simple array in which each interrupt can only have one handler. This is sensible, developers can just create a function that calls more than one handler if they need to call multiple functions when a specific interrupt happens.

asiekierka:
No, you never could assign more than one handler to each instruction - `irqSet` would overwrite matching masks, and `jump_intr` would never return to look for further handlers. The functionality is thus unchanged.

AntonioND:
Huh, then it was a weird design...

PR #106, opened 2024-06-15
dma: minor fixes

asiekierka:
- Added DMA_START_LINE definition.
- Documented DMA_START definitions.
- Fixed DMA_START defines on ARM7.
- glCallList() and cardStartTransfer() now use dmaSetParams().

### DIFF
--- a/include/nds/dma.h
+++ b/include/nds/dma.h
@@ -70,15 +70,20 @@ extern "C" {
 
 #ifdef ARM7
 #    define DMA_START_VBL   (1 << 28) ///< Auto-start DMA on vertical blank
+#    define DMA_START_SLOT1 (2 << 28) ///< Auto-start DMA on Slot-1 card transfer
 #    define DMA_START_CARD  (2 << 28) ///< Auto-start DMA on Slot-1 card transfer
+#    define DMA_START_WIFI  (3 << 28) ///< Auto-start DMA on NTR Wi-Fi interrupt request (DMA0/DMA2 only)
+#    define DMA_START_SLOT2 (3 << 28) ///< Auto-start DMA on Slot-2 interrupt request (DMA1/DMA3 only)
 #endif
 
 #ifdef ARM9
-#    define DMA_START_VBL   (1 << 27) ///< Auto-start DMA on vertical blank
-#    define DMA_START_HBL   (2 << 27) ///< Auto-start DMA on horizontal blank
-#    define DMA_START_LINE  (3 << 27) ///< Auto-start DMA on the beginning of each drawn scanline
+#    define DMA_START_VBL   (1 << 27) ///< Auto-start DMA on each frame's vertical blank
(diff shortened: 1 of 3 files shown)


include/nds/dma.h
```diff
@@ -66,18 +66,20 @@ extern "C" {
 #define DMA_BUSY            BIT(31)
 #define DMA_IRQ_REQ         BIT(30)
 
-#define DMA_START_NOW       0
-#define DMA_START_CARD      (5 << 27)
+#define DMA_START_NOW       0         ///< Start DMA immediately
 
 #ifdef ARM7
-#    define DMA_START_VBL   BIT(27)
+#    define DMA_START_VBL   (1 << 28) ///< Auto-start DMA on vertical blank
```
AntonioND, line 72: `1 << 27`?

AntonioND (changes requested):

include/nds/dma.h
asiekierka: Actually, no! Both melonDS and gbatek point out that ARM7 only uses bits 28-29 for its DMA targets.

asiekierka: (Though perhaps it may be a good idea to write a test for this, as such.)

include/nds/dma.h
```diff
@@ -66,18 +66,20 @@ extern "C" {
...
+#    define DMA_START_CARD  (2 << 28) ///< Auto-start DMA on Slot-1 card transfer
 #endif
```
AntonioND, line 74: Can you add defines for option 3?

> 3  DMA0/DMA2: Wireless interrupt, DMA1/DMA3: GBA Cartridge Slot

It could be nice so that we use it in DSWiFi or a successor.

include/nds/dma.h
```diff
@@ -66,18 +66,20 @@ extern "C" {
...
 
 #ifdef ARM9
-#    define DMA_START_HBL   BIT(28)
-#    define DMA_START_VBL   BIT(27)
-#    define DMA_START_FIFO  (7 << 27)
-#    define DMA_DISP_FIFO   (4 << 27)
+#    define DMA_START_VBL   (1 << 27) ///< Auto-start DMA on vertical blank
+#    define DMA_START_HBL   (2 << 27) ///< Auto-start DMA on horizontal blank
```
AntonioND, line 78: Maybe add the note here that there is in GBATEK that mentions "paused during V-Blank". It may be useful to be used as a quick reference.

include/nds/dma.h
asiekierka: Do we still want to use `CARD`/`CART` nomenclature going forward, or migrate towards `SLOT1`/`SLOT2`?

AntonioND: I think `SLOT2` is way more clear, no question about that. 😅

asiekierka: Should be fixed now.

AntonioND:
Thanks!

PR #107, opened 2024-06-20
console: clean up consoleLoadFont()

asiekierka:
- Remove the convertSingleColor option. This should be done by your preferred asset conversion tool, such as Grit; in addition, we migrated to a proper 1bpp font a while ago for libnds itself.
- Use BIOS bit unpack routines instead of a custom one to handle font decompression.
- Always initialize the palette when an user palette is not provided.
- Add partial support for 2bpp fonts.


AntonioND:
This is good, but it's a compatibility break, so it needs to appear in the changelog and the upgrade guide.

PR #108, opened 2024-06-20
decompress: move from arm9 to common

asiekierka:
This code is applicable to ARM9 and ARM7 environments, without much in the way of anything that depends on the ARM9 CPU.


AntonioND (commented):
We're really going to need an implementation of `sassert` for the ARM7... Maybe using the no$gba console...

PR #109, opened 2024-06-22
include: introduce BLOCKSDS_STRICT and <oldnames.h>

asiekierka:
The idea here is that a new define is added, `BLOCKSDS_STRICT`, of the value `(major_version * 10000 + minor_version * 100 + patch_version)` for a given version of BlocksDS.

* By default, BlocksDS templates come with `BLOCKSDS_STRICT` defined to the version they are copied from; for example, BlocksDS 1.3.0 templates will come with `-DBLOCKSDS_STRICT=10300` as a default.
* If defined to a given BlocksDS version, all deprecations made in that version or earlier are made inaccessible (no function, type definitions, etc).
    * For example, `IRQ_NETWORK` was renamed to `IRQ_RTC` in BlocksDS 0.3.0, so the former name is only accessible for `BLOCKSDS_STRICT < 300`.
* If defined to `1`, we try to keep maximum reasonable compatibility with the latest upstream version of libnds.
* If not defined (or defined to `0`), we try to also keep compatibility with older versions of libnds.
* In code completion environments (clangd, IntelliSense, etc), we try to avoid displaying the deprecated variants.

The idea is to allow BlocksDS to gradually clean up and improve the code/naming quality of the libraries while providing shims and aliases for old naming, thus not breaking older homebrew projects (as much).

My main concern is that this is somewhat too over-engineered, but in discussions we seem to agree that maintaining a friendly migration path for older projects is a notable goal given the current state of NDS homebrew.

An alternate idea pitched was to fork `libnds` into an alternate `libnds2`, which removes and cleans up all the legacy stuff in one go. My concern with that is that code quality is an ongoing process - in a few years, we may well decide that we need a `libnds3`.

### DIFF
--- a/include/nds/libversion.h
+++ b/include/nds/libversion.h
@@ -5,10 +5,18 @@
 #ifndef LIBNDS_NDS_LIBVERSION_H__
 #define LIBNDS_NDS_LIBVERSION_H__
 
+#define _BLOCKSDS_MAJOR_ 1
+#define _BLOCKSDS_MINOR_ 3
+#define _BLOCKSDS_PATCH_ 0
+
+#define _BLOCKSDS_STRICT_CURRENT_ (((_BLOCKSDS_MAJOR_) * 10000) + ((_BLOCKSDS_MINOR_) * 100) + (_BLOCKSDS_PATCH_))
+
+// The below defines mark the version of libnds with which BlocksDS tries to be compatible.
+
 #define _LIBNDS_MAJOR_ 1
 #define _LIBNDS_MINOR_ 8
-#define _LIBNDS_PATCH_ 0
+#define _LIBNDS_PATCH_ 2
 
-#define _LIBNDS_STRING "libnds release 1.8.0"
+#define _LIBNDS_STRING "BlocksDS " _BLOCKSDS_MAJOR_ "." _BLOCKSDS_MINOR_ "." _BLOCKSDS_PATCH_ " (like libnds " _LIBNDS_MAJOR_ "." _LIBNDS_MINOR_ "." _LIBNDS_PATCH_ ")"
 
 #endif // LIBNDS_NDS_LIBVERSION_H__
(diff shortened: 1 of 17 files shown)


include/nds/libversion.h
```diff
@@ -5,10 +5,18 @@
...
+
+// The below defines mark the version of libnds with which BlocksDS tries to be compatible.
+
 #define _LIBNDS_MAJOR_ 1
 #define _LIBNDS_MINOR_ 8
-#define _LIBNDS_PATCH_ 0
+#define _LIBNDS_PATCH_ 2
 
-#define _LIBNDS_STRING "libnds release 1.8.0"
+#define _LIBNDS_STRING "BlocksDS " _BLOCKSDS_MAJOR_ "." _BLOCKSDS_MINOR_ "." _BLOCKSDS_PATCH_ " (like libnds " _LIBNDS_MAJOR_ "." _LIBNDS_MINOR_ "." _LIBNDS_PATCH_ ")"
```
AntonioND, line 20: I'm not sure about this. If we're dropping compatibility with old stuff, we might as well drop the old version number. Also, if we wanted to do the same in libraries such as DSWiFi we would have two sets of `_BLOCKSDS_xxxx_`.

asiekierka: It's provided for compatibility reason - the idea, as we discussed, is that we're (mostly) keeping compatibility with libnds, no? Only when `_BLOCKSDS_STRICT_` is defined do we start breaking things en masse.

As for the other libraries: maxmod isn't likely to see major changes (its API is fairly clean and featureset largely complete), while DSWiFi has its own `dswifi_version.h` file and a much smaller API surface.

asiekierka: To clarify:

- I kept the integer version number so that defines which try to detect the presence of a specific libnds version continue working as expected,
- I changed the string mostly so that if someone uses it to print the version of libnds used by their code, it will be easy to distinguish between devkitARM's upstream and our own builds.

PR #110, opened 2024-07-06
interrupts: Avoid trying to read AUX IE, IF in NTR mode.

asiekierka:
We shouldn't be reading uninitialized memory on NTR.


asiekierka:
The way the previous interrupt dispatcher handled this is by only handling one IRQ at a time, then letting the interrupt handler be retriggered (as IF would remain != 0). This way, it didn't need to check `__dsimode` at all. As the "faster" approach now f.e. ignores the case where `IME = 0` and `IE != 0`, maybe that's the better way?

PR #111, opened 2024-07-06
interrupts: Do not retrigger self to fix uninitialized AUX IE/IF reads in NTR mode.

asiekierka:
Alternate approach to https://github.com/blocksds/libnds/pull/110 ; see that PR for notes.


AntonioND:
This approach looks better than #110, let's go with it.

PR #112, opened 2024-07-06
threads: add partial implementation of C11 threads API, wrapping cothreads

asiekierka:
It's very incomplete, as it only maps those functions of <threads.h> which our cothread/comutex system already implements.


AntonioND:
Looks pretty good! I guess we will find issues with it as we use it, but let's merge it for now.

PR #114, opened 2024-07-09
rsa: fix RSA functions paramaters

mdigiorgio:
The destination length in swiRSADecryptRAW needs to be a pointer a 32-bit unsigned integer.

The patch also provides documentation for functions, paramaters and structures.

### DIFF
--- a/include/nds/rsa.h
+++ b/include/nds/rsa.h
@@ -3,6 +3,12 @@
 //
 // Copyright (C) 2017 Dave Murphy (WinterMute)
 
+// DSi RSA functions
+
+/// @file nds/rsa.h
+///
+/// @brief DSi RSA functions.
+
 #ifndef LIBNDS_NDS_RSA_H__
 #define LIBNDS_NDS_RSA_H__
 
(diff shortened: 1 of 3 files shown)


include/nds/rsa.h
```diff
@@ -14,23 +20,77 @@ extern "C" {
...
+/// SHA1-HMAC value (RSA-SHA1).
+///
+/// @param ctx RSA Heap context
+/// @param dst Pointer to output buffer (should be 128 bytes).
+/// @param sig Pointer to signature buffer (should be 128 bytes).
+/// @param key Pointer to key buffer (should be 128 bytes).
+/// @return 1 if success, 0 if failed
 int swiRSADecrypt(swiRSAHeapContext_t *ctx, void *dst, const void *sig, const void *key);
+
+/// Decrypt signature data and do some extra processing for extracting a
```
asiekierka, line 72: This is copy-pasted material from GBATEK. As surprising as it may seem, GBATEK is not openly licensed (especially given at least its 3DS section is itself largely copy-pastes), and thus copying long snippets from it into BlocksDS is not acceptable.

include/nds/rsa.h
```diff
@@ -14,23 +20,77 @@ extern "C" {
...
 
-int swiRSAInitHeap(swiRSAHeapContext_t *ctx, void *heapStart, size_t heapSize);
-int swiRSADecryptRAW(swiRSAHeapContext_t *ctx, swiRSAbuffers_t *rsabuffers, size_t len);
+/// Initialize RSA Heap
+///
+/// @param ctx 12-byte heap information structure that gets set with heap
+///            start, heap end and heap length.
+/// @param heapStart Pointer to the start of the heap.
+/// @param heapSize Size of the heap in bytes (should be usually 4096).
+void swiRSAInitHeap(swiRSAHeapContext_t *ctx, void *heapStart, u32 heapSize);
```
asiekierka, line 45: Some suggestions here:

1. `4096` could probably be extracted into a constant, like `#define SWI_RSA_DEFAULT_HEAP_SIZE 4096`.
2. `swiRSAInitHeap`, to me, clearly exists to be plugged into an external allocator - which we do have, in the form of picolibc's `malloc()`. I would like to see a wrapper `swiRSAInit` which only takes a `heapSize` argument and handles allocation automatically (as well as a matching `swiRSAExit` which performs deallocation); in this variant, to further make things friendly, a `heapSize` of `0` would be treated as `SWI_RSA_DEFAULT_HEAP_SIZE`.

@AntonioND  ?

source/common/rsa.c
```diff
@@ -6,22 +6,21 @@
 #include <nds/rsa.h>
 #include <nds/system.h>
 
-int swiRSAInitHeapTWL(swiRSAHeapContext_t *ctx, void *heapStart, size_t heapSize);
```
AntonioND, line 9: I would keep the types as `size_t`. It's guaranteed to be 32-bit wide in the DS, and it's better semantically. A `u32` can be any unsigned value, a `size_t` represents a size.

include/nds/rsa.h
```diff
@@ -3,6 +3,12 @@
 //
 // Copyright (C) 2017 Dave Murphy (WinterMute)
 
+// DSi RSA functions
+
+/// @file nds/rsa.h
```
AntonioND, line 8: You will also need to go to `include/nds.h` and add a new entry for this file or it won't show up in the index.

include/nds/rsa.h
AntonioND: 1. Yeah, good idea.
2. We can add the wrapper later (once we have an example that actually works). We have a similar situation with the decompression routines, where we have the raw SWI calls and then wrapper functions around them.

include/nds/rsa.h
mdigiorgio: Sorry, forgot to check about license issues. Will fix

AntonioND:
It looks better now, can you squash all your commits manually and rebase them? I prefer to apply commits without the UI of GitHub to preserve dates and authors as in the original commit, but I don't want to merge your commits as a branch because the history still contains the GBATEK text.

mdigiorgio:
@AntonioND do you see more commits? I can only see one and my branch is rebased on top of latest master

source/common/rsa.c
```diff
@@ -6,22 +6,32 @@
...
 int swiRSADecryptPGPTWL(swiRSAHeapContext_t *ctx, void *dst, const void *sig, const void *key);
 
-int swiRSAInitHeap(swiRSAHeapContext_t *ctx, void *heapStart, size_t heapSize)
+swiRSAHeapContext_t createRSAHeapContext(size_t heapSize)
+{
+    swiRSAHeapContext_t rsaCtx;
+    rsaCtx.heapSize = (heapSize == 0) ? SWI_RSA_DEFAULT_HEAP_SIZE : heapSize;
+    rsaCtx.heapStart = malloc(rsaCtx.heapSize);
+    rsaCtx.heapEnd = ((uint8_t *)rsaCtx.heapStart) + rsaCtx.heapSize;
+    return rsaCtx;
```
AntonioND, line 22: I think that what asie meant was to create a function that is essentially:

```
int swiRSADecryptRAWTWL(swiRSAbuffers_t *rsabuffers, size_t *lenDst)
{
    swiRSAHeapContext_t rsaCtx = { 0 };
    rsaCtx.heapSize = ...
    ...

    return swiRSADecryptRAW(&rsaCtx, rsabuffers, lenDst)
}
```

Not a function that allocates the context.

AntonioND: Also, remember to check if `malloc()` returns `NULL`.

AntonioND:
> @AntonioND do you see more commits? I can only see one and my branch is rebased on top of latest master

Yes, sorry, I didn't realize you had done that.

mdigiorgio:
As discussed offline with @AntonioND , the wrappers need bit more thinking to decide the API to provide, so we'll do it at a later time.

AntonioND:
Let's go with this for now, yeah. Thanks!

PR #115, opened 2024-07-14
Define bitshift values for bitmasks

IvanVeloz:
On `memory.h` there are no bitshifts defined for the bitmasks, so it is impossible to shift a bit field back to the zero position using defined constants.

This is useful for addressing arrays for example (one example is an array of strings for debugging). There are hacks around it but this is a more proper solution.

Here is an example of what this change enables:

``` C

u16 exmemshadow = 0x00;
exmemshadow = EXMEMCNT_SRAM_TIME_8_CYCLES | EXMEMCNT_ROM_TIME1_8_CYCLES | EXMEMCNT_ROM_TIME2_6_CYCLES;
REG_EXTMEMCNT = exmemshadow; 

/* On a different function, we want to print the EXMEMCNT_ROM_TIME1 configuration. 
 * This is currently impossible without some hacks (division, case statements, etc.)
 */
char romtime1[4][10] = {"10 cycles", "8 cycles", "6 cycles", "18 cycles"};
int  romtime1_i = (REG_EXMEMCNT & EXMEMCNT_ROM_TIME1_MASK) >> EXMEMCNT_ROM_TIME1_SHIFT; // could be a macro
printf("ROM time 1 is %s", romtime1[romtime1_i]);

/* prints "ROM time 1 is 8 cycles". */
```

This is relevant because the main method of debugging on a phisical NDS is printf statements.

This also makes possible very neat macros ([e.g. libtonc's `tonc_core.h`](https://www.coranac.com/man/tonclib/group__grpCoreBit.htm)). These macro collections are common in embedded HALs. Porting libtonc's macros to BlocksDS could be a relatively easy way to get this functionality (of course, I still need to fix any other missing `_SHIFT` definitions on libnds).


AntonioND:
Thanks!

PR #116, opened 2024-07-14
LLVM Compatibility

GalaxyShard:
Fixes errors and warnings when compiling with LLVM.

This also changes the file extension of assembly files to `.S`, the standard for GCC and LLVM. For case-insensitive filesystems, `-x assembler-with-cpp` is specified as before.


Makefile.arm7
```diff
@@ -63,7 +63,7 @@ endif
 # Source files
 # ------------
 
-SOURCES_S	:= $(shell find -L $(SOURCEDIRS) -name "*.s")
+SOURCES_S	:= $(shell find -L $(SOURCEDIRS) -name "*.S")
```
AntonioND, line 66: This is a bad idea. Relying on the _case_ of the extension is going to cause issues for sure. We use lowercase extensions, but they should be considered case-insensitive.

I know that's how GCC does things, but that's a terrible idea, and it's not part of the C standard, as far as I know, so we should ignore that convention.

include/nds/arm9/math.h
```diff
@@ -129,7 +129,13 @@ static inline int32_t sqrtf32_result(void)
 }
 
 #pragma GCC diagnostic push
+// GCC does not recognize clang diagnostic pragma, ignore it here
```
AntonioND, line 132: Wouldn't it be better to do ``#ifdef __clang__`` or something like that?

source/arm7/tmio.twl.c
```diff
@@ -18,7 +18,7 @@
 #define INIT_DELAY_FUNC()  swiDelay(TMIO_CLK2DIV(400000u) * 74 / 4)
 
 
-static u32 g_status[2] = {0};
+static _Atomic u32 g_status[2] = {0};
```
AntonioND, line 21: Why is this needed?

AntonioND (changes requested):

Makefile.arm7
AntonioND: Also, if you have a repository on Windows and the _case_ of a file name changes, the actual file name won't change because Windows thinks they are the same name, so this isn't even going to work everywhere.

source/arm7/tmio.twl.c
GalaxyShard: The C11 standard requires variables to be marked as _Atomic in order to use atomic operations. GCC seems to have a "bug" of sorts that allows this to compile without marking it properly.

AntonioND: Sure, but what's the issue if this isn't marked as atomic?

GalaxyShard: statusPtr uses atomic operations so LLVM, following the standard, must have _Atomic on it. g_status needs it because statusPtr is a pointer to it, and _Atomic pointers are not compatible with the regular equivalent in GCC. Perhaps this could be solved with a cast, although I am unsure whether or not this could cause issues.
```c
source/arm7/tmio.twl.c: In function 'TMIO_sendCommand':
source/arm7/tmio.twl.c:254:36: error: initialization of '_Atomic u32 * const' {aka _Atomic long unsigned int * const'} from incompatible pointer type 'u32 *' {aka 'long unsigned int *'} [-Wincompatible-pointer-types]
  254 | 32 *const statusPtr = &g_status[controller];
      |                       ^
```

include/nds/arm9/math.h
GalaxyShard: Yea, not sure why I didn't think of that.

Makefile.arm7
GalaxyShard: If I'm not mistaken, searching for `.S` in a case-insensitive filesystem/OS should include `.s` files, so it should be fine on Windows. Now, it could cause issues if a `.s` file is made from windows and then compiled on Linux, so I think I'll just revert those changes keeping with the current lowercase convention.

include/nds/arm9/math.h
```diff
@@ -129,10 +129,11 @@ static inline int32_t sqrtf32_result(void)
 }
 
 #pragma GCC diagnostic push
```
AntonioND, line 131: I meant something like:
```c
#pragma GCC diagnostic push
#ifndef __clang__
#pragma GCC diagnostic ignored "-Wbuiltin-declaration-mismatch"
#endif
```

source/arm7/tmio.twl.c
AntonioND: In that case, you need to describe the problem a bit in the PR description.

include/nds/arm9/math.h
GalaxyShard: Updated.

source/arm7/tmio.twl.c
GalaxyShard: I clarified my comment & updated the description to reflect the current state of the PR.

AntonioND:
Oh, and I'm not saying that we need to implement that helper, I'm saying we probably want to see if we can get rid of the atomic instructions. Why are they needed?

GK-Gaming:
always_inline is removed everywhere on clang but,
doesnt clang support always inline?

https://clang.llvm.org/docs/AttributeReference.html#always-inline-force-inline

AntonioND:
I agree that we can remove the `always_inline` if it's purely based on thinking that the compiler will optimize the code. We should only use it if, for some reason, we need the function to always be inlined for other reason.

AntonioND:
BTW, now I remember that in this case the reason is that we can't have duplicated definitions of a function. In libnds currently the swi functions are duplicated, and the only thing that makes the compiler happy is that the definition of the header is never compiled as a regular function.

GalaxyShard:
> always_inline is removed everywhere on clang but, doesnt clang support always inline?

The definitions of those functions are replaced with declarations for the already-defined ones (in an assembly file) when using Clang, so theres nothing to inline from the header file.

GK-Gaming:
> BTW, now I remember that in this case the reason is that we can't have duplicated definitions of a function. In libnds currently the swi functions are duplicated, and the only thing that makes the compiler happy is that the definition of the header is never compiled as a regular function.

ahh ok.
also its fine i didnt see it was just the tiny bios call functions that were always_inine.
i thought it was more of the library

> The definitions of those functions are replaced with declarations for the already-defined ones (in an assembly file) when using Clang, so theres nothing to inline from the header file.

i see, i didnt notice that earlier

also, if you dont mind answering: why is there extra definition for gcc in header, but clang only needs declarations for the assembly file?

AntonioND:
> also, if you dont mind answering (dont want to spam this pr with random question and answers, etc...): why is there extra definition for gcc in header, but clang only needs declarations for the assembly file?

The `always_inline` version is just a way for the compiler to not require a branch to the function. It's a waste of a jump because the swi functions are just one instruction.

You could also get rid of the definitions in the header and it would work. However, this would mean that there are always jumps to the swi functions. Even link-time optimizations wouldn't work because the swi function is an assembly function, so it can't inline it.

source/arm7/tmio.twl.c
AntonioND: Okay, so this is causing build issues because of missing atomic builtins. It just won't build in GCC.

I think the right thing to do would be to get rid of all the atomic constructs and switch to using `volatile` if needed.

GalaxyShard:
I just compared four different `llvm-objdump -d` of tmio.twl.c.o built using GCC. The first one was built without this PR, the second was built with the defines below to replace atomic load/store, the third one was with built with both the defines and volatile added to statusPtr & g_status, and finally the last one was with this PR as it is now.
```c
#define GET_STATUS(ptr)       (*(ptr))
#define SET_STATUS(ptr, val)  (*(ptr) = (val))
```
Here are the objdump disassemblies
https://pastebin.com/hNsqLfAx - compiled with atomic load/store (before this pr)
https://pastebin.com/8KHLPtNU - compiled without atomic load/store (get/set_status replaced with the defines above)
https://pastebin.com/rzKSk1Hx - compiled with volatile & defines
https://pastebin.com/QUDsiaqe - compiled with _atomic (this pr, as of this comment; doesn't link correctly)

Looks like just replacing the defines adds 52 lines/instructions in the diff (which is a bit odd?), then adding volatile throws in another 20 lines on top of that.

GalaxyShard:
I found the 3DS version @profi200 mentioned in https://github.com/blocksds/sdk/issues/184#issuecomment-2226874344: 

https://github.com/profi200/libn3ds/blob/master/source/drivers/tmio.c

It looks like it uses `_Atomic u32` just like this PR.

profi200:
Yes, i added the atomic type later after i was informed of this problem. Also yeah, volatile is a very strong compiler barrier and will cause the compiler to optimize less leading to bigger code. Since we are using IRQ handlers here i found atomics to be an appropriate tool and it happens to optimize better as well.

If you use memory_order_relaxed there should not be any atomic library call on CPUs that don't support atomics in hardware unless clang is doing something weird.

source/arm7/tmio.twl.c
GalaxyShard: Latest commit should fix the linking errors; the SDK, examples, and templates now build on my machine with GCC, and libnds compiles with LLVM.

asiekierka:
So long as GCC performance doesn't regress, I'm fine with the PR.

Thank you for the elaborate explanation!

AntonioND (approved):

AntonioND:
Thanks!

PR #117, opened 2024-07-16
Make transfer region public again

AntonioND:
At some point in the early years of libnds, the transfer region was made private and all programs that used it stopped building. This was made in order to make people use the FIFO functions, which are more reliable.

However, even though it's very discouraged to use this, it can be useful to port old programs to the current version of libnds.

This is probably a polemic change but, honestly, I think it's a good idea. So far, I've been hardcoding the address of the transfer region structures of the games I've ported to BlocksDS (https://github.com/AntonioND/wolveslayer and https://github.com/AntonioND/talesofdagur) as a temporary step before migrating to using the FIFO, but porting old games is pretty hard, so it's still nice to have the option of delaying the migration from transfer region to FIFO (which is arguably the biggest change you have to make...).

### DIFF
--- a/include/nds.h
+++ b/include/nds.h
@@ -144,6 +144,7 @@ extern "C" {
 #include <nds/sha1.h>
 #include <nds/system.h>
 #include <nds/timers.h>
+#include <nds/transfer_region.h>
 #include <nds/touch.h>
 
 #ifdef ARM9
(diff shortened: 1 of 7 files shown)


AntonioND:
Yeah, fair enough. I've been thinking about it today and this is more trouble than it's worth.

PR #119, opened 2024-07-18
exceptions: Refactor exception and crash handling

AntonioND:
- A new exception handler has been added to libnds. This is considered a release exception handler which will print a short message such as "Data abort" or "Stack corrupted" so that the user knows there has been a critical error and they can tell the developer. This is equivalent to the situation when you are running a Linux application and it prints "segmentation fault" when it crashes.

  This exception handler doesn't use printf(), which makes the code footprint of this handler a lot smaller than the current debug handler.

- libnds now sets up the exception handler before main(). It's a bad idea to ask users to enable the exception handler themselves because they will forget.

- A new crash handler has been added. When libnds detects a fatal error it can't recover from (and which makes continuing unsafe) it will use this handler to print a short error message and let the user know.

- Error handling in the ARM7 is still quite bad, we need to think about how to communicate errors to the ARM9.

- Also, this PR makes the tone of red of the exception handler a bit darker so that it contrasts more with the text on the screen.

### DIFF
--- a/include/nds/arm9/exceptions.h
+++ b/include/nds/arm9/exceptions.h
@@ -46,8 +46,17 @@ void enterException(void);
 ///     Exception handler routine.
 void setExceptionHandler(VoidFn handler);
 
-/// Sets the default hardware exception handler.
+/// Sets the default debug hardware exception handler.
+///
+/// This handler prints a lot of information, like the state of the CPU
+/// registers when the CPU crashed.
 void defaultExceptionHandler(void);
+
+/// Sets the release hardware exception handler.
+///
+/// This is similar to defaultExceptionHandler(), but it only prints a minimal
+/// error message, and it uses a lot less code to do it.
+void releaseExceptionHandler(void);
 
 #ifdef __cplusplus
 }
--- a/source/arm7/libnds_internal.h
+++ b/source/arm7/libnds_internal.h
@@ -25,4 +25,13 @@ typedef struct {
  */
 libnds_touchMeasurementFilterResult libnds_touchMeasurementFilter(u16 values[5]);
 
+// In the ARM7 we can't really print anything without adding a lot of additional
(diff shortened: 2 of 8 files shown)


profi200:
I think the shorter exception message in release builds is not a good idea for 1 reason. Let's say a user reports a bug to you where it crashes. You can reproduce the exception in release mode but not in debug mode. Anabling the full handler in release mode also makes the exception go away. See what i mean? This scenario happens surprisingly often.

AntonioND:
The problem is that the default exception handler adds more code than the short one. It's not a huge difference, but it's a difference that every libnds program needs to pay.

Also, a super short error message is more useful than just a black screen or a freeze. We aren't trying to find a perfect way to debug programs, just something that is useful.

AntonioND:
I've rebased the branch and fixed a few conflicts. I've also started to use the pseudoinstruction `udf` instead of having a long message about how undefined instructions are formed.

I've been thinking about this for ages, so I'll probably go with the current PR unless someone complains in the next day or two.

I think this current solution is reasonable, even though it increases the binary size of everything. I care mostly about the fact that we need to print error messages from libnds. There are some errors that can't be returned to the user code to be handled. For example, if stack corruption is detected, it isn't even safe to return to user code at all. The only safe way to end is to do like Linux programs and print a message before exiting.

profi200:
It's rare but i have seen it happen. In case the exception handler itself crashes again i would add a lock to the exception handler that prevents running it again while in exception state.

AntonioND:
I was actually thinking about not using `printf()` in that handler at all, which is the only thing that can realistically cause an issue. All we would need is a function to print strings, and a function to print hex numbers.

AntonioND:
I've merged the PR with a few changes here: https://github.com/blocksds/libnds/commit/849171e2e60917a5ac2ece2511bad49a6bec6feb

PR #120, opened 2024-07-26
touch: Check the return value of cdcTouchReadData().

asiekierka:


AntonioND:
Thanks! This fixes the issue on my DSi and melonDS in DSi mode.

PR #124, opened 2024-08-09
sound: Add sound capture helpers

AntonioND:
The sound capture can be used to capture the output of sound channels 0 and 2, or the output of the sound mixer to the left or right speakers. The captured audio can be saved to RAM or added to the output of sound channels 1 and 3.

While audio capture isn't as useful as video capture, it's still useful to have a way for users to take advantage of it without too much effort.

Audio capture channel 0 requires sound channel 1 to be active. The sample rate of the capture is taken from the sample rate of playback of channel 1 (so that it can be added to it if required). The same applies to audio capture channel 1 and sound channel 2.

The helpers in this patch only let the user setup the capture registers. Sound channels 1 and 3 must be setup by hand. The user may want to play the captured audio there, or it may want to silence it (by setting the volume to 0) and only use the values saved to RAM.

### DIFF
--- a/include/nds/arm7/audio.h
+++ b/include/nds/arm7/audio.h
@@ -62,6 +62,23 @@ extern "C" {
 #define REG_SNDCAP0CNT      (*(vu8 *)0x04000508)
 #define REG_SNDCAP1CNT      (*(vu8 *)0x04000509)
 
+#define SND0CAPCNT_CH1_OUT_DIRECT       (0 << 0)
+#define SND0CAPCNT_CH1_OUT_ADD_TO_CH0   (1 << 0)
+#define SND0CAPCNT_SOURCE_LEFT_MIXER    (0 << 1)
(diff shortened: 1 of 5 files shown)


AntonioND:
There is an example of this in the following PR: https://github.com/blocksds/sdk/pull/195

I'm not super convinced of how the usage looks like. I thought about being able to provide the sample rate in `soundCaptureStart()`, but that would make it a bit tricky to setup channel 1/3 if you want to play it at the same time as audio is being captured.

AntonioND:
If you need more information, you can check GBATEK: https://problemkaputt.de/gbatek.htm#dssoundcapture

AntonioND:
Ah, turns out that you don't even need to enable an audio channel for the timer value to work:
```
    soundSetFreq(1, CAPTURE_SAMPLE_RATE);
    soundSetFreq(3, CAPTURE_SAMPLE_RATE);
```
This works just as well, and it's pretty clean. I will finish the documentation of this PR and merge it.

PR #126, opened 2024-09-01
nitrofs: Block direct Slot-2 access for >32MB .nds files

asiekierka:
As a Slot-2 cartridge can have a maximum of 32MB of mapped address space, any .nds file over 32MB is guaranteed not to be fully accessible via this method. In particular, NO$GBA mounts the first 32MB of a .nds file in Slot-2, which caused NitroFS reads over the 32MB limit to be corrupted.


AntonioND:
Thanks!

PR #127, opened 2024-09-01
include, libc: fix includes for latest picolibc commits

asiekierka:
Note that this works with both current (in-repo) and latest (will be pushed with the next BlocksDS update) picolibc version.


AntonioND:
Thanks!

PR #130, opened 2024-09-28
cardEeprom, ndsmotion: clean up, fix bugs

asiekierka:
- Use AUXSPICNT macros instead of magic numbers for card SPI operations.
- Use Kionix macros instead of magic numbers for DS Motion Card SPI operations.
- Add motion_get_type() and motion_get_name() to the motion sensor driver.
- Clean up code duplication in the motion sensor driver.
- Fix motion sensor driver bugs:
  - Potentially fix DS Motion Card support in DSi mode.
  - Fix unnecessary DS card writes on deinitialization.
  - Fix analog input access causing unwanted GBA I/O when a DS card is detected.

The DS Motion Card proper may now work in DSi mode, but I don't have one to test; I have tested the MK6/R6, but that one seems to not communicate properly with a DSi/3DS at all.

### DIFF
--- a/include/nds.h
+++ b/include/nds.h
@@ -164,6 +164,7 @@ extern "C" {
 #    include <nds/arm9/keyboard.h>
 #    include <nds/arm9/linkedlist.h>
 #    include <nds/arm9/math.h>
+#    include <nds/arm9/ndsmotion.h>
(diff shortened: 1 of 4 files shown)


AntonioND:
Thanks!

PR #132, opened 2024-10-04
LLVM and undefined behavior fixes

GalaxyShard:
* Add an immediate value (0) to `udf` instructions, as it is required for LLVM and optional with GCC
* Fix signed overflow from `1 << 31`
* Proper printf format string in gurumeditation (fixes warnings)
* Change an implicit enum cast to explicit, with a comment explaining why it is valid

The only other error/warning still in LLVM 19 is the following:

```c
libnds/source/common/libc/sbrk.c:43:20: error: variable 'stack_ptr' is uninitialized when used here
        heap_end = stack_ptr;
                   ^
libnds/source/common/libc/sbrk.c:25:33: note: initialize the variable 'stack_ptr' to silence this warning
    register uintptr_t stack_ptr asm("sp");
                                ^
```

It seems like GCC's `register` keyword is explicitly not intended to be used the way it is in sbrk.

From https://gcc.gnu.org/onlinedocs/gcc/Local-Register-Variables.html:

> The only supported use for this feature is to specify registers for input and output operands when calling Extended asm

It could potentially be replaced with an inline assembly `mov` like this:

```diff
diff --git a/source/common/libc/sbrk.c b/source/common/libc/sbrk.c
index bb818ed..fc75feb 100644
--- a/source/common/libc/sbrk.c
+++ b/source/common/libc/sbrk.c
@@ -22,7 +22,8 @@ void *sbrk(int incr)
     const uintptr_t end = (uintptr_t)__end__;
 
     // Trick to get the current stack pointer
-    register uintptr_t stack_ptr asm("sp");
+    uintptr_t stack_ptr;
+    asm ("mov %0, sp" : "=r" (stack_ptr) );
 
     // Next address to be used. It is updated after every call to sbrk()
     static uintptr_t heap_start = 0;
```

However I am not sure of the semantics are correct, and GCC produces different code, so I didn't include it as part of this PR.

### DIFF
--- a/source/arm9/system/gurumeditation.c
+++ b/source/arm9/system/gurumeditation.c
@@ -441,7 +441,7 @@ void libndsCrash(const char *msg)
     exceptionMsg = msg;
 
     // Use an undefined instruction defined by the assembler
-    asm volatile("udf" ::: "memory");
+    asm volatile("udf #0" ::: "memory");
 
     while (1);
 }
--- a/source/arm9/system/gurumeditation.c
+++ b/source/arm9/system/gurumeditation.c
@@ -6,6 +6,7 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <inttypes.h>
 
 #include <nds/arm9/background.h>
 #include <nds/arm9/console.h>
@@ -338,7 +339,7 @@ void guruMeditationDump(void)
     {
         // Finally, print everything to the screen
 
-        printf("  pc: %08lX addr: %08lX\n\n", codeAddress, exceptionAddress);
+        printf("  pc: %08" PRIX32 " addr: %08" PRIX32 "\n\n", codeAddress, exceptionAddress);
 
         for (int i = 0; i < 8; i++)
         {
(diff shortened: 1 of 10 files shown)


AntonioND:
Looks reasonable. Can you rename the commit titles of `fix missing immediate on udf instruction` and `fix undefined behavior caused by 1 << 31` to have the prefix `chore: `?

source/arm9/system/gurumeditation.c
```diff
@@ -338,7 +339,7 @@ void guruMeditationDump(void)
     {
         // Finally, print everything to the screen
 
-        printf("  pc: %08lX addr: %08lX\n\n", codeAddress, exceptionAddress);
+        printf("  pc: %08"PRIX32" addr: %08"PRIX32"\n\n", codeAddress, exceptionAddress);
```
AntonioND, line 342: Can you leave spaces between the double quotes and the `PRIX32` defines?

GalaxyShard:
Updated commit messages and added spacing around PRIX32

AntonioND:
Thanks!

PR #134, opened 2024-10-30
Changes to square root functions

Kuratius:
Changes square root register types to unsigned.
Changes square root helper arguments and return types to unsigned. 
Starts integer square root earlier on repeated calls.

I'm unsure about the second point because it seems to me that it's at least partially intended to use these functions with a signed fixed point type, but it doesn't do any checks for negative numbers and instead just treats it as an unsigned number anyway so having them as a signed type isn't useful.

### DIFF
--- a/include/nds/arm9/math.h
+++ b/include/nds/arm9/math.h
@@ -9,6 +9,7 @@
 ///
 /// @brief hardware coprocessor math instructions.
 
+
 #ifndef LIBNDS_NDS_ARM9_MATH_H__
 #define LIBNDS_NDS_ARM9_MATH_H__
 
@@ -122,7 +123,7 @@ static inline int32_t mulf32(int32_t a, int32_t b)
 ///     20.12 positive value.
 static inline void sqrtf32_asynch(uint32_t a)
 {
-    REG_SQRT_PARAM = ((uint64_t)) << 12;
+    REG_SQRT_PARAM = ((uint64_t)a) << 12;
 
     if ((REG_SQRTCNT & SQRT_MODE_MASK) != SQRT_64)
         REG_SQRTCNT = SQRT_64;
@@ -404,8 +405,6 @@ static inline uint32_t sqrt64(uint64_t a)
 ///
 /// @warning
 ///     Not safe to call inside an interrupt handler.
-/// @warning
-///     Cannot be used concurrently with integer square root functions.
 ///
 /// @param x
 ///     Valid 32 bit non-negative floating point value.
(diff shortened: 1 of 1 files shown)


include/nds/arm9/math.h
```diff
@@ -120,9 +120,9 @@ static inline int32_t mulf32(int32_t a, int32_t b)
 ///
 /// @param a
 ///     20.12 positive value.
-static inline void sqrtf32_asynch(int32_t a)
+static inline void sqrtf32_asynch(uint32_t a)
 {
-    REG_SQRT_PARAM = ((uint64_t)(uint32_t)a) << 12;
+    REG_SQRT_PARAM = ((uint64_t)) << 12;
```
AntonioND, line 125: Please, at least test that the code builds before sending a PR.

include/nds/arm9/math.h
```diff
@@ -400,6 +404,8 @@ static inline uint32_t sqrt64(uint64_t a)
 ///
 /// @warning
 ///     Not safe to call inside an interrupt handler.
+/// @warning
+///     Cannot be used concurrently with integer square root functions.
```
AntonioND, line 408: Does this warning really matter? If anything, we should have it at the top of the file, not in one function (and leave the rest unchanged).

AntonioND (changes requested):

include/nds/arm9/math.h
Kuratius: The idea was to highlight a behavior difference to normal sqrtf. Someone might assume this is some extra function instead of repurposing the existing functions.

include/nds/arm9/math.h
Kuratius: I changed it to a draft for now.

include/nds/arm9/math.h
AntonioND: Yeah, but the div functions have the same issue, right? That's why I mentioned that a note on top of the file would be nice:

```
/// @brief hardware coprocessor math instructions.
///
/// Warning: there is only one divider and one square root unit [...]
```

include/nds/arm9/math.h
```diff
@@ -9,6 +9,7 @@
 ///
 /// @brief hardware coprocessor math instructions.
 
+
```
AntonioND, line 12: Revert this newline.

include/nds/arm9/math.h
Kuratius: I added a warning to the top. Let me know if it looks right in doxygen.

Kuratius:
I also wrote a modified version of your hw_math test that could be used for speed testing the integer implementations. Let me know if I should open a PR for it.

AntonioND:
Sure, send a PR with the changes to the tests.

PR #135, opened 2024-10-31
ndsmotion: Implement support for the retail Motion Pack (ATTiny-based)

asiekierka:
Fixes https://github.com/blocksds/sdk/issues/216 . Requires real hardware validation.


asiekierka:
Validated both DSi support and retail Motion Pack support. Special thanks to @beta215 for testing like a dozen builds on real hardware.

AntonioND:
Thanks!

PR #136, opened 2024-11-03
cache, diskio: Optimizations and bugfixes.

asiekierka:
- Fix cache invalidation not invalidating more than one sector per call.
- Replaced cache algorithm with a simple LRU; appears to perform better with simpler code, but more benchmarking/research should be done.
- Removed the malloc() call in disk_write(), replacing it with "borrowing" a slot from the LRU cache without marking it valid.
- Adjusted disk_read() similarly, so that sectors marked as not to be cached do not generate cache requires.
- Added additional debugging defines to diskio.c.

### DIFF
--- a/source/arm9/libc/fatfs/cache.c
+++ b/source/arm9/libc/fatfs/cache.c
@@ -13,7 +13,7 @@ typedef struct
     uint8_t  valid;
     uint8_t  pdrv;
     LBA_t    sector;
-    uint32_t usage_count;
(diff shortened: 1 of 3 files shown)


asiekierka:
For future reference: the benefits of replacing the old cache code with a simple LRU are uncertain; however, the previous cache code didn't pass scrutiny as well, and we lack benchmarks for realistic real-world patterns.

The remaining changes seem to hold up nicely.

PR #137, opened 2024-11-04
mpu, vectorbase: Protect the alternate vector base region.

asiekierka:
This PR disables writes to the alternate vector base region (0x00000000) in all cases. It also disables reads to the region if the alternate vector base is not opted into. This helps detect null pointer access/write attempts in homebrew, as these will now raise an exception.


include/nds/system.h
```diff
@@ -221,14 +221,14 @@ static inline void systemShutDown(void)
...
-/// @note
-///     ARM9 only
-void setVectorBase(int highVector);
+/// Set the ARM9 interrupt vector base.
+///
+/// @param base
+///     Vector base. Setting it to any non-zero value will use the vector base
+//      at 0xFFFF0000 (provided by the BIOS); setting it to zero will use the
+//      vector base at 0x00000000 (provided by the homebrew program,
+//      initialized with BIOS pointers by default - see SystemVectors).
```
AntonioND, line 230: This is unsafe, isn't it? You can't write the interrupt handler to the vector memory until you have called `setVectorBase()` but, at that point, any interrupt will jump there, even before you can write your handler. This needs to have a `@warning` saying that `REG_IME` should be set to 0 before calling it, the interrupt handler needs to be copied right away, and only then it is safe to enable interrupts by setting `REG_IME` to 1. Am I missing something?

asiekierka: > You can't write the interrupt handler to the vector memory until you have called `setVectorBase()`

You can. The writes happen to `0x01000000`; notice that `setVectorBase()` never enables write permissions to the region.

AntonioND: Ah, right, I forgot about that. Then add a comment to the function to explain that `0x01000000` is a mirror of `0x00000000` and that users should copy their handlers there before calling `.setVectorBase()`.

asiekierka:
```
//      vector base at 0x00000000 (provided by the homebrew program,
//      initialized with BIOS pointers by default - see SystemVectors).
```

This already points the user to the structure they're supposed to write the pointers to their memory handlers into. The implementation detail of them actually being written to at `0x01000000` is irrelevant at this level; it is documented in `mpu_setup.s`.

Either way, I have clarified the documentation further, and fixed some mistakes.

AntonioND:
Thanks!

PR #138, opened 2024-11-05
exceptions: Expose custom exception handler functionality to ARM7

asiekierka:
Initial stepping stone for https://github.com/blocksds/sdk/issues/177 , should you choose to implement it; however, note that I haven't tested it (would probably be fairly easy to write a test using an `udf` on the ARM7).


AntonioND:
I'll try to test this tonight.

AntonioND:
Okay, I don't have time to write a test for this right now. If someone decides to use it and it's broken, just fill an issue.

PR #145, opened 2024-11-27
Let glColorTableEXT allocate VRAM without filling

ds-sloth:
We briefly discussed this over Discord around an hour ago.

This change is necessary to allow allocating a palette while the 3D engine is active without causing VRAM corruption.

I've been using a similar change (as a separate function built with my program) successfully for months, and have tested this change as performing equivalently to my earlier version.


AntonioND:
Thanks!

PR #146, opened 2024-11-28
cp15: optimize CP15_CleanAndFlushDCache

asiekierka:
Appears to save up to ~5-6%. Can be benchmarked with https://github.com/asiekierka/nds-misc-tests/tree/main/cached-memory-performance

Not sure if this routine should be in ITCM or not, haven't tested the consequences - likewise the `cache.c` equivalent.

### DIFF
--- a/source/arm9/system/cp15.s
+++ b/source/arm9/system/cp15.s
@@ -273,29 +273,26 @@ BEGIN_ASM_FUNC CP15_SetITCM
     bx      lr
 
 //////////////////////////////////////////////////////////////////////
-// Routine obtained from page 3-11 of ARM DDI 0201D
+// Routine adapted from page 3-11 of ARM DDI 0201D
 //////////////////////////////////////////////////////////////////////
 BEGIN_ASM_FUNC CP15_CleanAndFlushDCache
 
-    // Loop in all 4 segments
-    mov     r1, #0
-outer_loop:
-
-    // Loop in all entries in one segment
+    // Loop in all entries in all 4 segments
     mov     r0, #0
-inner_loop:
-    orr     r2, r1, r0 // Generate segment and line address
-    mcr     CP15_REG7_CLEAN_FLUSH_DCACHE_ENTRY_BY_INDEX(r2) // Clean the line
+    // r1 adjusts r0 to be set to the beginning of the next segment after completed iteration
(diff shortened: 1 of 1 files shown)


source/arm9/system/cp15.s
```diff
@@ -273,26 +273,23 @@ BEGIN_ASM_FUNC CP15_SetITCM
...
+    beq     .Linner_loop
 
-    add     r1, r1, #0x40000000 // Increment segment counter
-    cmp     r1, #0x0
-    bne     outer_loop
+    adds    r0, r0, r1 // Increment segment counter
+    bne     .Linner_loop
 
     // Drain write buffer
     mov     r0, #0
```
profi200, line 295: Not needed because r0 will already be 0 after the above loops.

asiekierka: Already done independently; forgot about it the first time around. Thanks!

lifehackerhansol:
Sorry for hijacking the PR a bit; apparently the behaviour of the write buffer drain is wrong, according to section 4.8 of [ARM946-PRDC-000592 5.0](https://documentation-service.arm.com/static/5ed4c52cca06a95ce53f91ae?token=)

From that doc it seems the drain should happen immediately before every flush in inner_loop.

Figured it'd be worth pinging here as the function in question is being modified in this PR.

profi200:
The document is not exactly clear on whenever this is needed for both range and full cache operations. The examples only show the address variants (used in range flush). In this PR only the full cache flush got modified.

asiekierka:
Pushed a proposal of handling the errata; however, it would be useful, albeit not mandatory, to validate that the errata happens on NTR consoles in the wild in these specific conditions. Luckily, Jakly is working on emulating ARM9 cache timings, so maybe that will happen...

AntonioND:
I'm going to close this for now. I assume that ARM made sure that the routine in the official manual was correct. Unless it's clear that the alternative routine is safe, it's better to use the official one. Feel free to reopen it if I'm wrong!

asiekierka:
The errata may post-date the official manual. Either way, I think a reproduction is necessary first...

profi200:
Zero issues so far with the optimized ones. I don't believe there is a special reason why the example code from ARM isn't optimal but i agree that the errata needs to be tested.

PR #147, opened 2024-12-21
Add ARM7 exception handler that sends the crash dump to the ARM9

AntonioND:
This PR adds an exception handler that the user needs to enable. When the exception handler is executed, it saves a crash dump in the IPC shared memory and sends a FIFO message to the ARM9 to notify it so that it can print it (the same way it prints ARM9 crash dumps). Note that the FIFO message is just one packet that is sent without using the FIFO library so that the process is more robust.

It also splits the ARM9 exception handler so that it can reuse the printing code for ARM9 and ARM7 crashes.

### DIFF
--- a/include/nds/exceptions.h
+++ b/include/nds/exceptions.h
@@ -54,7 +54,6 @@ void enterException(void);
 ///     Exception handler routine.
 void setExceptionHandler(VoidFn handler);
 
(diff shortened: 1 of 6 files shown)


AntonioND:
You can use this example to verify it: https://github.com/blocksds/sdk/pull/243

PR #148, opened 2024-12-24
Add initial version of the ARM7 console

AntonioND:
The ARM9 allocates a ring buffer in main RAM. This buffer is passed to the ARM7, which then can start writing text to it. The ARM7 needs to notify the ARM9 when new text is added to the buffer so that the ARM9 can print it and empty the buffer.

The system is connected to the current ARM9 console system, so the output of this console can be any PrintConsole in use by the ARM9. For example, it is possible to start one console on the top screen and let the ARM9 use it while there is another console on the bottom screen that is used to print whatever the ARM7 prints.

A custom versions of `puts()` has been implemented, as well as a minimal version of printf(). ARM7 binaries don't have a lot of space to play with, so it's a generally a bad idea to use the regular printf() there.

It's still possible to use snprintf() from libc instead of this minimal version of printf() if more complex formats are required.

### DIFF
--- a/include/nds.h
+++ b/include/nds.h
@@ -99,6 +99,7 @@
 ///
 /// @section arm7 ARM7 modules
 /// - @ref nds/arm7/clock.h "RTC utilities"
+/// - @ref nds/arm7/console.h "Console to that sends messages to the ARM9 easily"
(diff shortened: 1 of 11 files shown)


AntonioND:
Example here: https://github.com/blocksds/sdk/pull/246

PR #149, opened 2024-12-28
bios.h: add constants for TUnpackStruct flag

GalaxyShard:
https://github.com/blocksds/sdk/pull/249


AntonioND:
Can you rename the commit to just `bios:`, not `bios.h:`?

You need to add it to the doxygen documentation too:
```
/// This is a description
#define DEFINE_1 1
/// Help for another define
#define DEFINE_2 2
```

Also, you should mention the defines from the documentation of field `dataOffset` of `TUnpackStruct`.

AntonioND:
Thanks!

PR #151, opened 2024-12-31
peripherals: fix compile bug in rumble.c

asiekierka:


AntonioND:
Thanks!

source/arm9/peripherals/rumble.c
```diff
@@ -31,7 +31,7 @@ bool rumbleIsEdgeActivated(void)
 
 #define RUMBLE_PAK_CTRL (*(vuint16 *)0x08001000)
 #define GPIO_DATA       (*(vuint16 *)0x080000C4)
-extern void slot2EzCommand(uint32_t address, uint16_t value);
+extern void __libnds_slot2EzCommand(uint32_t address, uint16_t value);
```
GalaxyShard, line 34: I think technically a double underscore prefixing an identifier is undefined behavior (C23 section 6.4.2.1), but then again this naming scheme is also used for __libnds_exit and I don't think it's caused any issues (maybe if more warnings are enabled?)

AntonioND: Yeah, you're right, actually. Maybe we could have a PR to fix all libnds internal names with that issue.

PR #152, opened 2025-01-02
console: include used headers directly

GalaxyShard:
`sys/cdefs.h` is what defines `__printflike`, currently console.h relies on cdefs.h being included beforehand.


davidgfnet:
Saw the same issue in my Fedora/Ubuntu packages (https://github.com/davidgfnet/blocksds-toolchain)
Thanks!

AntonioND:
That's what I was going to say, you were faster than me!

Thanks for the fix!

PR #153, opened 2025-01-08
bios: Use custom version of swiSoftReset() instead of SWI #0

AntonioND:
This SWI isn't available on DSi, only DS, so it's better to not rely on it at all and use the version that libnds has in source/common/swiSoftReset.s.

### DIFF
--- a/include/nds/bios.h
+++ b/include/nds/bios.h
@@ -121,17 +121,11 @@ typedef struct UnpackStruct
 #define BIOS_UNPACK_OFFSET_ADD_NON_ZERO (0u << 31)
 
 /// Resets the DS.
-#ifdef __clang__
+///
+/// Note that this function doesn't actually call the BIOS with a SWI
+/// instruction. SWI #0 (SoftReset) isn't available on DSi, only on DS.
 __attribute__((noreturn))
 void swiSoftReset(void);
-#else
-__attribute__((always_inline, noreturn))
-static inline void swiSoftReset(void)
-{
-    asm volatile inline ("swi 0x0 << ((1f - . == 4) * -16); 1:");
-    __builtin_unreachable();
-}
-#endif
 
 /// Delays the code.
 ///


asiekierka:
I'm not sure if continuing to call libnds's replacements for BIOS functions `swiX` is a good idea - it's part of why we got into this mess to begin with, by the distinction not being apparent.

AntonioND:
Yeah, that's fair enough. I'm going to merge my currently active PR and then add a layer of indirection to all the custom SWIs.

AntonioND:
Or... not. If I rename the functions, any ASM code calling them will break. Unless I add wrappers that are callable from ARM (not static inline functions in a header), but then we're back to the initial situation.

AntonioND:
https://github.com/blocksds/libnds/commit/dc1541aada60c4482a891c8b06fb32030b5b22eb

asiekierka:
> If I rename the functions, any ASM code calling them will break.

The idea I had was to add new `systemSoftReset` functions, keep `swiSoftReset` as BIOS wrappers, and mark them with `__attribute__((deprecated))` for C callers + document the distinction in the migration/upgrade guides. The same would be done for `intrWait` and `swiIntrWait` (https://github.com/blocksds/sdk/issues/27)

AntonioND:
I'm not sure. Marking them as deprecated would imply that at some point they will be removed. And removing some functions just so that maintainers of libnds don't get confused... I think documenting which functions are special is good enough.

I don't think this trick is particularly bad, anyway. The whole point why we are tricking users with the name is that originally the functions actually used the SWI instruction https://github.com/AntonioND/ndslib-archive/blob/eafb2703db6e2b6b414a2fa15107212522666bff/lib/ARM9/biosCalls.S#L54-L58 but then bugs were discovered in the BIOS and they were transparently replaced so that users wouldn't have to worry about changing their code: https://github.com/blocksds/libnds/commit/1c964c74b4e1f88bc50cba7f75004bf2d0f3bbc4

PR #154, opened 2025-01-08
bios: Add new swiIntrWaitAUX() for the ARM7

AntonioND:
The old prototype of swiIntrWait() doesn't support the extended interrupts of the ARM7. The prototype of swiIntrWaitAUX() supports the new flags.

However, the old prototype has been kept for compatibility with older code and so that code shared between ARM9 and ARM7 can use the same function.

### DIFF
--- a/include/nds/interrupts.h
+++ b/include/nds/interrupts.h
@@ -256,6 +256,23 @@ void irqDisableAUX(u32 irq);
 ///     implemented in libnds.
 void swiIntrWait(u32 waitForSet, uint32_t flags);
 
+#ifdef ARM7
+/// Wait for interrupt(s) to occur. DSi ARM7 only.
+///
+/// @param waitForSet
+///     0: Return if the interrupt has already occured; 1: Wait until the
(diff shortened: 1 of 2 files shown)


include/nds/interrupts.h
```diff
@@ -253,6 +253,9 @@ void irqDisableAUX(u32 irq);
 ///     This doesn't actually use a software interrupt, it's a custom function
 ///     implemented in libnds.
 void swiIntrWait(u32 waitForSet, uint32_t flags);
+#ifdef ARM7
```
AntonioND, line 256: This needs to be documented (well, and the other ARM7 functions of this file, they are ignored by doxygen because the description only affects one function, not two).

AntonioND:
This needs to be tested.

include/nds/interrupts.h
AntonioND: https://github.com/blocksds/libnds/commit/78f1a5e760904a86ebecf928e222eee62a5a01d0

PR #155, opened 2025-01-16
fatfs: Update to R0.15ap1

asiekierka:


AntonioND:
Thanks for the update!

PR #156, opened 2025-01-19
Hardware-accelerated inverse floating point square root

Kuratius:
https://github.com/Aikku93 wrote a software divide that can keep up with the square root unit and wrote this implementation in assembly. 
They asked me to make a pull request for it. Draft for now because I need to update the hw_math test also.

Some general information on the characteristics of this function:
Execution time seems to be about 37 bus cycles on dsi, 61-62 bus cycles on ds.
Accuracy should be within 1 ulp of the result you would get when doing the calculation with doubles and then casting to single precision, which should more accurate than a simple 1.0f/sqrtf(x). 
The result is exact (well, identical) in ~89 % of possible mantissas.

Considerations: In this version the LUT is in main memory, but it may be desirable to place it in tcm on ds. I believe on dsi it doesnt matter where you put it because the overall execution time is limited by the hw square root. Alignment to cache lines could also be a consideration.

### DIFF
--- a/include/nds/arm9/math.h
+++ b/include/nds/arm9/math.h
@@ -415,6 +415,21 @@ static inline uint32_t sqrt64(uint64_t a)
 ///     32 bit floating point value
 ARM_CODE float hw_sqrtf(float x);
 
+/// 32-bit inverse floating point sqrt
+///
+/// @warning
+///     Subnormal floats are treated as if they were zero.
+///
+/// @warning
+///     Not safe to call inside an interrupt handler.
+///
+/// @param x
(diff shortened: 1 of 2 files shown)


Kuratius:
Apologies, this was not intended to two commits.

Kuratius:
There's potentially a bug in the Inf or NaN branch; I'll add a test for it  to the other PR and check if the fix suggested by aikku works.

Kuratius:
Closing, as I am working on a version that doesn't require the HW accelerator, so this would end up redundant. I'll post updates in the issue on the main repo if there is anything interesting.

PR #157, opened 2025-02-14
palette connectCount check

sillysagiri:
when attempting to unload a texture where the palette is assigned from another texture using glAssignColorTable(), the program crashes.

it seems that there is error when removing palette from texture. palette->connectCount should be checked if its `<= 0` instead

this resolve the issue


AntonioND:
Thanks! Can you just rename the commit to `videoGL: Fix palette connectCount check`?

sillysagiri:
done

AntonioND:
Thanks!

PR #160, opened 2025-03-09
fatfs: Minor improvements and bugfixes

asiekierka:
Found while researching adding file system re-initialization, but these seem valuable enough to PR them a bit early.


AntonioND:
Thanks!

PR #162, opened 2025-04-14
Camera improv

Kuratius:
Needs someone to test on 3DS.

### DIFF
--- a/source/arm9/peripherals/camera.twl.c
+++ b/source/arm9/peripherals/camera.twl.c
@@ -37,6 +37,7 @@ static bool updateDmaParam(uint16_t width, uint16_t height){
     if (scanlines>16){
         return false;
     }
+    //note: wlength % blength should be 0, i.e. total number of words should be an integer multiple of the words in a logical block
     camera_state.dma_wlength=total_words; //total number of words per transfer
     camera_state.dma_blength=(scanlines*bytes_per_scanline+4-1)/4; //number of words until bdelay pause happens to give arm9 space to execute
     camera_state.dma_scanlines=scanlines; //number of scanlines per logical block
--- a/source/arm9/peripherals/camera.twl.c
+++ b/source/arm9/peripherals/camera.twl.c
@@ -56,7 +56,11 @@ bool cameraSetCaptureModeTWL(u8 captureMode)
     fifoWaitValue32(FIFO_CAMERA);
     bool ret=fifoGetValue32(FIFO_CAMERA);
     fifoMutexRelease(FIFO_CAMERA);
-    if (!ret) return ret;
+    if (!ret) 
+    {
+        camera_state.last_mode=-1; //if an error happened we dont know what our capture mode is
+        return ret;
+    }
     camera_state.last_mode=captureMode;
     return true;
 }
--- a/source/arm9/peripherals/camera.twl.c
+++ b/source/arm9/peripherals/camera.twl.c
@@ -15,14 +15,16 @@ extern u8 cameraActiveDevice;
 
 static struct camera_state {
     int8_t last_mode;//<- actually used
-    uint8_t  dma_scanlines; //<-these extra params are for future custom resolutions
+/*  uint8_t  dma_scanlines; //<-these extra params are for future custom resolutions
     uint32_t dma_wlength; // after an access to last_mode they should be in cache
     uint32_t dma_blength; //
     uint16_t height;//
     uint16_t width;//
+*/
 } camera_state;
 
 //This function is for when support for different resolutions gets added to the camera driver
+/*
 static bool updateDmaParam(uint16_t width, uint16_t height){
     //camera buffer is 512 words
     if (width==0 || height==0)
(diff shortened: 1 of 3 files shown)


source/arm9/peripherals/camera.twl.c
```diff
@@ -13,6 +13,49 @@
...
+    int8_t last_mode;
+    uint8_t  dma_scanlines; //future use
+    uint32_t dma_wlength; //
+    uint32_t dma_blength; //might be doable as u16, but better safe than sorry
+    uint16_t height;//
+    uint16_t width;//
+} __camera_state;
+
+//for future use
+bool updateDmaParam(uint16_t width, uint16_t height){
```
asiekierka, line 26: If it's not publicly exposed, make the function `static`.

source/arm9/peripherals/camera.twl.c
```diff
@@ -13,6 +13,49 @@
 
 extern u8 cameraActiveDevice;
 
+struct camera_state {
+    int8_t last_mode;
+    uint8_t  dma_scanlines; //future use
```
asiekierka, line 18: `future use` doesn't tell anything to the next developer who looks at this code; and in open source, code changes hands *a lot*. Same for the comment below.

source/arm9/peripherals/camera.twl.c
```diff
@@ -13,6 +13,49 @@
...
+    uint32_t dma_wlength; //
+    uint32_t dma_blength; //might be doable as u16, but better safe than sorry
+    uint16_t height;//
+    uint16_t width;//
+} __camera_state;
```
asiekierka, line 23: Same ofr this - if it's not used outside of `camera.twl.c`, it is `static`.

AntonioND: Also, names that start with two underscores are reserved (and names with one underscore followed by an uppercase letter).

Basically, this struct is only seen inside of this C file, so the name of the struct can be anything as long as it's `static`.

source/arm9/peripherals/camera.twl.c
Kuratius: I added some more clarification. My thinking was that calculating the correct dma parameters is annoying, so a helper somewhere would be welcome.

source/arm9/peripherals/camera.twl.c
```diff
@@ -13,6 +13,49 @@
...
 extern u8 cameraActiveDevice;
 
+static struct camera_state {
+    int8_t last_mode;//<- actually used
+    uint8_t  dma_scanlines; //<-these extra params are for future custom resolutions
+    uint32_t dma_wlength; // after an access to last_mode they should be in cache
+    uint32_t dma_blength; //
+    uint16_t height;//
+    uint16_t width;//
+} camera_state;
```
AntonioND, line 23: This struct is just taking up a lot of RAM when all it needs is a byte to store the last mode.

source/arm9/peripherals/camera.twl.c
```diff
@@ -13,6 +13,49 @@
...
+    uint32_t bytes_per_scanline=(uint32_t)width*bytes_per_pixel;
+    uint32_t total_bytes=bytes_per_scanline*height;
+    uint32_t total_words=(total_bytes+4-1)/4;
+    uint32_t scanlines=(512*4)/(bytes_per_scanline);
+    if (scanlines>16){
+        return false;
+    }
+    camera_state.dma_wlength=total_words; //total number of words per transfer
+    camera_state.dma_blength=(scanlines*bytes_per_scanline+4-1)/4; //total number of words per camera interrupt
+    camera_state.dma_scanlines=scanlines; //number of scanlines after which camera interrupt happens
```
AntonioND, line 40: I would leave this as a comment:
```
// DMA words length = total_words
// DMA bytes (not sure what you meant by "b") length = ...
// DMA scanlines = ...
```

source/arm9/peripherals/camera.twl.c
```diff
@@ -91,38 +135,39 @@ bool cameraSelectTWL(CameraDevice device)
     }
 }
 
-bool cameraStartTransferTWL(u16 *buffer, u8 captureMode, u8 ndmaId)
+void cameraStartDMA(u16 * buffer, u8 captureMode, u8 ndmaId)
```
AntonioND, line 138: This function should be `static`.

source/arm9/peripherals/camera.twl.c
Kuratius: We could just comment out the other variables for now.

source/arm9/peripherals/camera.twl.c
Kuratius: Blength is what the variable is called in the DMA defines. I think it's the number of words per logical block, not the number of bytes.

source/arm9/peripherals/camera.twl.c
Kuratius: Do think this applies to start capture too or just the DMA start function?

source/arm9/peripherals/camera.twl.c
Kuratius: It pauses dma for 2 timer ticks after every logical block, to give the arm9 some space to breathe.

source/arm9/peripherals/camera.twl.c
AntonioND: Well, any function that can be useful needs to be in a public header. Functions in this file are supposed to be loaded only in DSi mode, so you will need to wrap them and call them from `camera.c`, and then add them to `camera.h` with the right documentation.

Also, I'm thinking that functions in this file should have the `TWL` suffix so that it's easier to know that they are in `.twl` sections.

source/arm9/peripherals/camera.twl.c
AntonioND: Ah, I see. Well, in any case, if nothing is using it, it shouldn't be taking up RAM.

PR #163, opened 2025-04-21
include: Prepare for picolibc 1.8.10 update

asiekierka:
Remove a picolibc-ism that has been cleaned up in picolibc 1.8.10, instead use the equivalent GCC attribute.

This shouldn't break building with picolibc 1.8.9, so it can be merged sooner.


AntonioND:
Thanks!

PR #165, opened 2025-05-19
math: fix normalization of large and small fixed point vectors

Kuratius:
This technically computes extra digits that are shifted out at the end so you have the option of further increasing the accuracy by 1 ulp.
It can no longer be inlined, but any possible gain from inlining is outweighed by the use of arm mode and using the division accelerator less.

There's also a minor optimization to hw_sqrtf.

### DIFF
--- a/include/dlfcn.h
+++ b/include/dlfcn.h
@@ -96,6 +96,27 @@ char *dlerror(void);
 ///     error.
 void *dlsym(void *handle, const char *name);
 
+/// Returns a pointer to the base address of the code loaded with the library.
+///
+/// The purpose of this function is to print this address to be used when
+/// debugging the code with emulators. You must provide this address when
+/// loading the elf file of the library. For example, with the GDB command line:
+///
+/// Note that this is a libnds-specific function.
+///
+/// ```
+/// add-symbol-file path/to/dynamic.elf -s .progbits <load_address>
+/// ```
+///
+/// @param handle
+///     The handle returned by dlopen().
+///
+/// @return
+///     On success, it returns a pointer to the base address of the code loaded
+///     to RAM. On error it returns NULL, and the user is expected to call
+///     dlerror() to get a user-readable string with the reason of the error.
+void *dlmembase(void *handle);
+
 #ifdef __cplusplus
 }
 #endif
--- a/source/arm9/dlfcn.c
+++ b/source/arm9/dlfcn.c
@@ -703,3 +703,19 @@ void *dlsym(void *handle, const char *name)
     dl_err_str = "symbol not found";
     return NULL;
 }
+
+void *dlmembase(void *handle)
+{
+    // Clear error string
+    dl_err_str = NULL;
+
+    if (handle == NULL)
+    {
+        dl_err_str = "invalid handle";
+        return NULL;
+    }
+
+    dsl_handle *h = handle;
+
+    return h->loaded_mem;
+}


Kuratius:
I'll see what speed I can get out of a fully optimized float version, I think the casts on the one I tried previously are very inefficient.

Kuratius:
This current one is about 87 ticks, 126 ticks ds
Optimized float version w/o hw is 62 ticks DSi, ds untested
I have an idea for using  the 32-by-32  mode but I will probably have to redo the derivation to get the shifts correct.
If using the 32-bit mode is possible then maybe 80 ticks is possible for HW accelerated on DS.

Kuratius:
I think it'll probably round wrong in some cases with only 16 bits on the division, this is probably the best possible balance of accuracy and speed.

Kuratius:
Also, the compiler has a tendency to generate very weird assembly for the very last shift (nonsense like generating the same instruction with the same registers and opposite condition codes). That can be fixed with extra inline asm but it adds extra complexity and I'm not sure it's worth it. The compiler also tends not to use stm/ldm for the last section but I'm not sure if that's worth it generally as the write buffer and cache exist so using ldm is probably not that critical.

AntonioND:
First, I'd split the changes to the current code from the improvements to the normalization function.

Honestly, I'm not too sure about this. This code seems to be optimized for one specific compiler, and it looks like it may completely change if a different compiler version decides to do things in a different way.

Kuratius:
I just noticed the original function also isn't compiled as ARM, changing that  and testing on a dsi gets me 78 ticks for the version in this PR and 186 for the default version. Still quite a bit faster but only 2x and not 3x like I originally claimed.

AntonioND:
I've merged your first commit, with the improvements to `hw_sqrtf()`. You can rebase your branch on top of master.

The thing is that I agree this is a nice performance improvement to have, so the new crazy code is tempting.

Let's do something. We can keep the old code and the new one just in case, and as reference.
```
#if 0
// Old code
#else
// New code
#endif
```
Also, remember to add a comment to the Doxygen description of the normalize function to warn users that this uses the hardware accelerators.

Kuratius:
I applied a small fix to at least fix the worst of the overflow problems by not truncating intermediate results, the performance of the commented out version is the same as the previous implementation.

Kuratius:
Also, because I had a thought about it: the noinline tag may be removable if you edit the inline asm to use a generic register rather than r1,r2,r3 (i.e. letting the compiler choose the register). If you think that would be better then that can be adjusted.
If the compiler decides to inline it into a thumb project it will however override the ARM_CODE tag and destroy advantage of inlining.

Kuratius:
At least in melonDS it actually benches slightly faster with generic registers chosen by the compiler, on hardware I saw no difference.

Kuratius:
I got some ideas for optimizing it further from people on discord like avoiding an extra stall cycle, but I think this is probably already about as good as it gets without compromising readibility or writing everything in asm.

source/arm9/math.c
```diff
@@ -105,3 +105,77 @@ ARM_CODE float hw_sqrtf(float x)
...
+    {
+        //a[i] = divf32(a[i], magnitude);
+        start_div64_64((int64_t)a[i]<<12, (int64_t) magnitude);
+        a[i]=finish_div64_64();
+    }
+}
+#endif
+
+#if 1
+__attribute__((noinline)) ARM_CODE void normalizef32(int32_t *a)
```
AntonioND, line 123: You don't need this attribute here.

source/arm9/math.c
```diff
@@ -105,3 +105,77 @@ ARM_CODE float hw_sqrtf(float x)
...
+{
+    register int32_t a1;
+    register int32_t a2;
+    register int32_t a3;
+    register uint32_t Hi;
+    register uint32_t Lo;
+    //This assembly statement may give a warning about wrong register order for ldm
+    //ignore it, for the compiler doesn't know addition of squares will commute
```
AntonioND, line 131: I'm not a fan of ignoring compiler warnings, to be honest. If there is a way to make it go away, I think it's better.

AntonioND (commented):
I think it would be better if you didn't use the same PR for multiple things. You should have a PR for the normalization optimizations, and another one for the new division function. If not, I keep seeing updates to this PR and I never know if you're done with it or not. And then I check again and there are a lot of new unrelated changes...

Kuratius:
I added the new functions because you suggested keeping the old code around and I thought you wanted to re-enable it for debugging or in case the the new code breaks something. The thing is that the old code is fundamentally broken, if it worked for anyone before it was probably due to luck, so I think it shouldn't be left in this state. Performance is a secondary concern in light of that.

source/arm9/math.c
Kuratius: It's possible to just declare r1,r2,r3 as clobbers and use them directly. It messes up a bit in later parts of the code because it seemingly has a strong preferences for using low registers in certain parts of the code, I assume it's probably thumb and arm code generation rules getting mixed up, or it's trying to keep some registers free to use for instructions that need neighboring registers or ascending registers, but it's not a big difference.

PR #166, opened 2025-05-29
exceptions: fix non-literal format string warning

GalaxyShard:
Fixes the following error generated by Clang/Zig:

```c
libnds/source/arm9/system/exceptions.c:32:12: error: format string is not a string literal (potentially insecure)
    printf(title);
           ^
libnds/source/arm9/system/exceptions.c:32:12: note: treat the string as an argument to avoid this
    printf(title);
           ^
```


AntonioND:
Thanks!

PR #169, opened 2025-06-28
math: correct semantics of asm in normalizef32()

19tracks:
Kuratius in the gbadev Discord noticed that this asm statement was missing a memory input constraint, since it reads from the memory pointed to by `a`.

I took a look at it, too, and noticed a couple more things. `volatile` isn't necessary because the statement could be eliminated if its outputs were never used, and the `&` modifiers aren't necessary because the inputs aren't read from after the outputs are written to. I also changed the input's name to match the name of the variable instead of the name of the register it will be kept in.

None of this affects the generated code with the current toolchain, but it is a little more correct.

### DIFF
--- a/source/arm9/math.c
+++ b/source/arm9/math.c
@@ -111,14 +111,14 @@ ARM_CODE void normalizef32(int32_t *a)
     register uint32_t Lo;
     register uint32_t Hi;
 
-    asm volatile (
+    asm (
         ".syntax unified \n\t"
-        "ldm %[r0], {r1,r2,r3} \n\t"
+        "ldm %[a], {r1,r2,r3} \n\t"
         "smull %[Lo],%[Hi], r1, r1 \n\t"
         "smlal %[Lo],%[Hi], r2, r2 \n\t"
         "smlal %[Lo],%[Hi], r3, r3 \n\t" :
-        [Lo]"=r&"(Lo), [Hi]"=r&"(Hi) :
-        [r0]"r"(a) :
+        [Lo]"=r"(Lo), [Hi]"=r"(Hi) :
+        [a]"r"(a), "m"(*(int32_t (*)[3]) a) :
         "r1", "r2", "r3"
     );
 


Kuratius:
I think having the & constraint on outputs is generally good practice in multi-line asm statements because they may want to use an input more often than in the first line.

If we're removing them, might be good to document it. Either here or in a guide future guide to inline asm for DS. 
"Normally we might need a "&" qualifier on the output but since the input is only needed for the first instruction and the clobbers already aren't allowed to overlap with inputs/outputs, it's not necessary." to help people who might try to emulate the code after reading it. 

Removing the & constraint could actually allow better code generation if you were to make this a separate function for use in other contexts where the input address isn't needed after computing the magnitude. That's also something that could be discussed.

19tracks:
I'll step through the changes with reference to the GCC documentation:

```
asm (
```

> [(Link)](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Volatile) GCC’s optimizers sometimes discard asm statements if they determine there is no need for the output variables. Also, the optimizers may move code out of loops if they believe that the code will always return the same result (i.e. none of its input values change between calls). Using the volatile qualifier disables these optimizations.

This asm statement has no effect other than to write to its outputs and clobbers, and it will produce the same outputs given the same inputs (the same pointer and the same pointed-to data), so `volatile` isn't needed, though leaving it wouldn't make the code incorrect.

```
    ...
    [Lo]"=r"(Lo), [Hi]"=r"(Hi) :
```

> [(Link)](https://gcc.gnu.org/onlinedocs/gcc/Modifiers.html#index-earlyclobber-operand) ‘&’ Means (in a particular alternative) that this operand is an earlyclobber operand, which is written before the instruction is finished using the input operands. Therefore, this operand may not lie in a register that is read by the instruction or as part of any memory address.

This asm statement has (after this change) two input operands: a register operand `a` and an unnamed memory operand. Both are used in the first instruction (ldm) and then never again. This ldm doesn't write to `Lo` or `Hi` (and ldm reads its input before writing its outputs anyway), so the earlyclobber modifier isn't necessary for either output, though leaving those wouldn't make the code incorrect, either.

(This asm statement doesn't use alternatives, so the parenthetical about them isn't relevant.)

```
    [a]"r"(a), "m"(*(int32_t (*)[3]) a) :
    ...
```

> [(Link)](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Output-Operands) Operands are separated by commas. Each operand has this format:
>
> `[ [asmSymbolicName] ] constraint (cvariablename)`
>
> asmSymbolicName
>
> Specifies an optional symbolic name for the operand. The literal square brackets ‘[]’ around the asmSymbolicName are required both in the operand specification and references to the operand in the assembler template, i.e. ‘%[Value]’. The scope of the name is the asm statement that contains the definition. Any valid C identifier is acceptable, including names already defined in the surrounding code. No two operands within the same asm statement can use the same symbolic name.

Changing the symbolic name from "r0" to "a" has no effect, even though the old name happened to be the same as the name of a register.

> [(Link)](https://gcc.gnu.org/onlinedocs/gcc/Simple-Constraints.html#index-memory-references-in-constraints) ‘m’ A memory operand is allowed, with any kind of address that the machine supports in general. Note that the letter used for the general memory constraint can be re-defined by a back end using the TARGET_MEM_CONSTRAINT macro.

This is how the "m" constraint is documented, but it tells us very little.

> [(Link)](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Clobbers-and-Scratch-Registers-1) Here is a fictitious sum of squares instruction, that takes two pointers to floating point values in memory and produces a floating point register output. Notice that x, and y both appear twice in the asm parameters, once to specify memory accessed, and once to specify a base register used by the asm. You won’t normally be wasting a register by doing this as GCC can use the same register for both purposes. However, it would be foolish to use both %1 and %3 for x in this asm and expect them to be the same. In fact, %3 may well not be a register. It might be a symbolic memory reference to the object pointed to by x.
>
> ```
>asm ("sumsq %0, %1, %2"
>     : "+f" (result)
>     : "r" (x), "r" (y), "m" (*x), "m" (*y));
> ```
>
> Here is a fictitious `*z++ = *x++ * *y++` instruction. Notice that the x, y and z pointer registers must be specified as input/output because the asm modifies them.
>
> ```
>asm ("vecmul %0, %1, %2"
>     : "+r" (z), "+r" (x), "+r" (y), "=m" (*z)
>     : "m" (*x), "m" (*y));
> ```
>
> An x86 example where the string memory argument is of unknown length.
>
> ```
>asm("repne scasb"
>    : "=c" (count), "+D" (p)
>    : "m" (*(const char (*)[]) p), "0" (-1), "a" (0));
> ```
>
> If you know the above will only be reading a ten byte array then you could instead use a memory input like: `"m" (*(const char (*)[10]) p)`.

This is buried in the "clobbers and scratch registers" section for some reason, but it tells us a lot more about how "m" constrained operands are used to tell the compiler which memory the assembly code may access. The code is incorrect without this change, but a more conservative/less precise way to make it correct would be to add a "memory" clobber.

Does that help give you confidence that the PR is correct? If you're still not sure, the least precise and most difficult to get wrong solution would be to leave `volatile` and the earlyclobber modifiers in place and just add a "memory" clobber instead of an "m" operand.

AntonioND:
That's a good explanation, thanks for taking the time to write it!

I'll merge the PR now, I'm finally free after dealing with the issues with the FIFO code.

AntonioND:
Hmmm GitHub did the rebase but the long description in your PR was missed, so I've added it manually: https://github.com/blocksds/libnds/commit/3481924b01c828598293c84468612306bf8e03d8

PR #170, opened 2025-07-07
WIP atan2 and 20.12 quotient helper

Kuratius:


Kuratius:
Right now the assembly routine has rounding, the C one doesn't.
One idea that is floating around in my head is to return a non standard float as a 64-bit value, where the lower 32-bits are the mantissa, and the high 32-bits are the exponent.
Getting the intended value would be trivial, as it would just be a lsr 
```(uint64_t)(uint32_t)retval >>((uint32_t)(retval>>32))

Then someone could decide if they want rounding, a float type, etc etc

PR #172, opened 2025-07-16
Ensure errno is reset in scandir so that readdir loop doesn't break early

shinyquagsire23:
Works, but someone should probably go through the POSIX spec or other impls to see if this is the way to go tbh

### DIFF
--- a/source/arm9/libc/scandir.c
+++ b/source/arm9/libc/scandir.c
@@ -34,6 +34,9 @@ int scandir(const char *path, struct dirent ***names,
     if (dir == NULL)
         return -1;
 
+    // ensure errno is reset so readdir loop doesn't break early
+    errno = 0;
+
     *names = NULL;
     while ((ent = readdir(dir)) != NULL)
     {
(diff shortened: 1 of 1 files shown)


Kuratius:
Might be better to temporarily save the current value of errno and restore it if the loop finishes without error, and otherwise overwrite it with the error.

shinyquagsire23:
@Kuratius thank you for that, I knew there probably had to be some POSIX lore so I'll go ahead and store/zero/restore then

source/arm9/libc/scandir.c
```diff
@@ -82,5 +89,11 @@ int scandir(const char *path, struct dirent ***names,
     }
 
     closedir(dir);
+
+    // restore previous errno if no new errnos
+    if (!errno) {
+        errno = errno_prev;
+    }
+
     return error ? -1 : count;
```
Kuratius, line 98: If we reach the return statement and error is set, do we also need to set errno to some value or is that handled by the functions we called?

shinyquagsire23: Yeah my thought is that if readdir or anything else errnos, it should be what gets seen by the application since that's the more immediate error.

PR #174, opened 2025-07-22
fatfs: Update to R0.16

asiekierka:
May require some testing, as it just came out today. You may also need to update your forked repository.


AntonioND:
Thanks! I agree this needs a bit of testing. I've updated the `fatfs` fork in the BlocksDS organization, but that won't have any effect until this PR is merged.

Also, I've tagged all versions of `wf-fatfs` ever used in BlocksDS. I realized it wouldn't be possible to go back in time from a fresh clone. Luckily, my local clone still had all references to the commits that I needed to tag: https://github.com/blocksds/wf-fatfs/tags

AntonioND:
This seems to be working, thanks!

PR #180, opened 2025-08-11
slot2: Distinguish between edge-activated and non-edge-activated rumble carts

asiekierka:
Non-edge-activated rumble can be driven with a bit less CPU/timer logic. The EZ Flash's status here is unknown.


AntonioND:
I've added your PR description to the commit and merged it here https://github.com/blocksds/libnds/commit/06ec20304544d69f5f13ad9c999a5fba050378ec

Thanks!

PR #181, opened 2025-08-23
math: use 64-bit intermediate for dotf32 and crossf32

Kuratius:
Except for INT32_MIN*1 inputs this increases accuracy in all cases. It also improves performance and code size due to using smlal.

c.f. https://godbolt.org/z/of6E8h3qn

Out of these two, crossf32 is the one where INT32_MIN can be problematic. 
Edit: I cant seem to trigger this issue for some reason, but I believe since the bit pattern for the additive inverse of INT32_MIN is identical to (uint32_t)INT32_MIN I think it should be an issue for smull.

### DIFF
--- a/include/nds/arm9/math.h
+++ b/include/nds/arm9/math.h
@@ -451,9 +451,9 @@ ARM_CODE float hw_sqrtf(float x);
 ///     Result pointer to fixed 3x3 matrix
 static inline void crossf32(int32_t *a, int32_t *b, int32_t *result)
 {
-    result[0] = mulf32(a[1], b[2]) - mulf32(b[1], a[2]);
-    result[1] = mulf32(a[2], b[0]) - mulf32(b[2], a[0]);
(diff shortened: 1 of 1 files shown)


Kuratius:
In thumb mode this is probably slightly slower than the current ones, but I'm also fairly certain arm mode is smaller than thumb mode here.. That said, this isn't super critical. There's some extra cases where this should provide a valid result where the previous ones didn't (like say the cross product of two extremely long vectors that point *almost* into the same direction, since for cross product
```
|a x b| = |a| * |b| * sin(theta)
``` and the sine would implicitly fix any overflows that would otherwise be cut off by truncating before adding.
Or the dot product of two extremely large vectors that are almost but not quite at a right angle.

If there was some way to tell the compiler to build these in arm mode I would, but I think the inline tag overrides the arm_mode tag.

Essentially what this change does is allow the compiler to use a fused multiply add instead of separate multiply and add. This reduces code size for arm mode builds.

PR #183, opened 2025-08-28
math: Add restrict qualifier to result in crossf32

Kuratius:
This allows tighter code generation in some cases.

https://godbolt.org/z/bnqnzobPf

Edit: Apparently thumb gets larger though.

Size in arm:
24 instructions (96 bytes) with restrict
35 instructions (140 bytes) without restrict

Size in thumb:
72 instructions without restrict (144 bytes)
97 instructions with restrict (196 bytes)

### DIFF
--- a/include/nds/arm9/math.h
+++ b/include/nds/arm9/math.h
@@ -449,7 +449,7 @@ ARM_CODE float hw_sqrtf(float x);
 ///     Pointer to fixed 3 dimensions vector.
 /// @param result
 ///     Pointer to fixed 3 dimensions vector that will contain the result.
-static inline void crossf32(const int32_t *a, const int32_t *b, int32_t *result)
+__attribute__((always_inline)) static inline void crossf32(const int32_t *a, const int32_t *b, int32_t *result)
 {
     if(__builtin_constant_p(a[0]) && __builtin_constant_p(b[0]) &&
        __builtin_constant_p(a[1]) && __builtin_constant_p(b[1]) &&
--- a/include/nds/arm9/math.h
+++ b/include/nds/arm9/math.h
@@ -449,7 +449,7 @@ ARM_CODE float hw_sqrtf(float x);
 ///     Pointer to fixed 3 dimensions vector.
 /// @param result
 ///     Pointer to fixed 3 dimensions vector that will contain the result.
-void crossf32(int32_t *a, int32_t *b, int32_t *result);
+void crossf32(const int32_t *a, const int32_t *b, int32_t *result);
 
 /// 20.12 fixed point dot product.
 ///
--- a/include/nds/arm9/math.h
+++ b/include/nds/arm9/math.h
@@ -449,7 +449,7 @@ ARM_CODE float hw_sqrtf(float x);
 ///     Pointer to fixed 3 dimensions vector.
 /// @param result
 ///     Pointer to fixed 3 dimensions vector that will contain the result.
-static inline void crossf32(int32_t *a, int32_t *b, int32_t *result)
+static inline void crossf32(int32_t *a, int32_t *b, int32_t *restrict result)
 {
     result[0] = ((int64_t)a[1] * b[2] + (int64_t)-b[1] * a[2]) >> 12;
     result[1] = ((int64_t)a[2] * b[0] + (int64_t)-b[2] * a[0]) >> 12;
--- a/include/nds/arm9/math.h
+++ b/include/nds/arm9/math.h
@@ -449,12 +449,7 @@ ARM_CODE float hw_sqrtf(float x);
 ///     Pointer to fixed 3 dimensions vector.
 /// @param result
 ///     Pointer to fixed 3 dimensions vector that will contain the result.
-static inline void crossf32(int32_t *a, int32_t *b, int32_t *restrict result)
-{
-    result[0] = ((int64_t)a[1] * b[2] + (int64_t)-b[1] * a[2]) >> 12;
-    result[1] = ((int64_t)a[2] * b[0] + (int64_t)-b[2] * a[0]) >> 12;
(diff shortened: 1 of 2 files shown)


Kuratius:
Gericom suggested that it might be better to have a and b be loaded into a temporary buffer first, so that overlapping with result is a safe thing to do. I'm uncertain if that's better or worse.

Kuratius:
Probably, it depends on how far the compiler looks into the function and if you enable lto.
With lto it should still be able to propagate a constant as long as the implementation is in C and not in assembly.
If it's usually used with normalizef32 then normalizef32 will already stop constant propagation.

Kuratius:
I tested it with a separate cross.c file in the math example in blocksDS.

```#include <stdint.h>



void cross(const int32_t *a,const int32_t *b, int32_t *result)
{
    int32_t ta[3]={a[0],a[1],a[2]};
    int32_t tb[3]={b[0],b[1],b[2]};
    result[0] = ((int64_t)ta[1] * tb[2] + (int64_t)-tb[1] * ta[2]) >> 12;
    result[1] = ((int64_t)ta[2] * tb[0] + (int64_t)-tb[2] * ta[0]) >> 12;
    result[2] = ((int64_t)ta[0] * tb[1] + (int64_t)-tb[0] * ta[1]) >> 12;
}
```

With no LTO and -O2, it doesn't constant propagate a function from a separate file into main.c. __builtin_constant_p on the result returns false.
With LTO and -O2, it correctly constant propagates. __builtin_constant_p on the result returns true.

-O3 has the same behavior.

source/arm9/math.c
```diff
@@ -172,11 +172,37 @@ ARM_CODE void normalizef32(int32_t *a)
     }
 }
 
-ARM_CODE void crossf32(const int32_t *a,const int32_t *b, int32_t *result)
+ARM_CODE void asm_crossf32(const int32_t *a,const int32_t * b,int32_t *result)
```
asiekierka, line 175: `libnds_` prefix for internal symbols please

Kuratius:
I know Antonio probably wants the asm in a separate file, but this should be good enough to discuss it at least.
It can also still be fully inlined if lto is activated.
I think the same trick could be added to normalizef32 to make it compile-time evaluatable if that is desirable.

asiekierka:
Note that I didn't necessarily say it *should* be compile-time evaluatable. I also was thinking about cases where, say, only some of the values are compile-time evaluatable.

While I admire the level of effort, it may be impractical in practice...

Kuratius:
You can change the && to a || if that is better I think?

source/arm9/math.c
```diff
@@ -171,3 +171,38 @@ ARM_CODE void normalizef32(int32_t *a)
         a[i] = prod;
     }
 }
+
+ARM_CODE void libnds_asm_crossf32(const int32_t *a,const int32_t * b,int32_t *result)
```
AntonioND, line 175: No, if the function is fully in assembly it should go into a different file that is just in assembly. This will only make it harder to understand because of the inline asm syntax. `%[r0]` isn't something I want to read unless strictly needed.

include/nds/arm9/math.h
```diff
@@ -449,11 +449,26 @@ ARM_CODE float hw_sqrtf(float x);
...
+    {
+        int32_t ta[3]={a[0],a[1],a[2]};
+        int32_t tb[3]={b[0],b[1],b[2]};
+        tb[1]=-tb[1];
+        result[0] = ((int64_t)ta[1]*tb[2] + (int64_t)tb[1]*ta[2])>>12;
+        ta[0]=-ta[0];
+        result[1] = ((int64_t)ta[2]*tb[0] + (int64_t)tb[2]*ta[0])>>12;
+        tb[0]=-tb[0];
+        result[2] = ((int64_t)ta[0]*tb[1] + (int64_t)tb[0]*ta[1])>>12;
+        //the weird minus signs are to keep it fully identical to the asm code
```
AntonioND, line 466: Why do we need it to be fully identical? It makes the code less readable, and we should focus on that, I think.

Kuratius: (These are without inlining)

source/arm9/math.c
Kuratius: I was too much mental overhead for me to think of a filename. If the asm implementation is too hard to read then maybe just using one of the C ones and ignoring inlining is best. Someone who needs inlining can select lto, anyway. Then you dont have an implementation in two different places.

include/nds/arm9/math.h
Kuratius: I think this would also work: 
```
void weird_crossf32(const int32_t *a, const int32_t *b, int32_t *restrict result)
{
        result[0] = ((int64_t)a[1]*b[2] + (int64_t)-b[1]*a[2])>>12;
        result[1] = ((int64_t)a[2]*b[0] + (int64_t)b[2]*-a[0])>>12;
        result[2] = ((int64_t)-a[0]*-b[1] + (int64_t)b[0]*a[1])>>12;
        //the weird minus signs are to keep it fully identical to the asm code
        return;
}
``` is that clearer?

Kuratius:
Typo in the last line of the commit message, 1 sec.

source/arm9/math.c
Kuratius: Another reason against both inline asm and asm might be that this functions reads and writes memory, with GCC doesnt handle correctly with address sanitizers, and gcc doesnt handle inline asm correctly with address sanitizers even if you give it the right constraints:
https://godbolt.org/z/vYEbTeYaq

For normalizef32 this problem is mitigated by C code again accessing the same memory later and the asm doing only a read, but for this asm would break sanitizers. So I'd assume assembly only functions should either be to fixed addresses like with hw dividers, or they shouldnt use memory beyond the obvious, like stack.

PR #190, opened 2025-09-12
Add support for mounting nand2 partition on DSi

edo9300:
This commits enables mutli partition support in fatfs, allowing the 2nd partition of the DSi nand to be mounted, with the same functions used for the primary nand partition.
This new partition will be accessible through the `nand2:/` drive.
On top of that, to avoid wasting memory in DS mode, the control structures for the DSi specific partitions (`sd`, `nand` and `nand2`) are now put into the twl memory region, saving 1,5kb of memory


AntonioND:
Thanks!

